PR #6, opened 2021-05-06
feat: plumb LD_DEV_EMAIL env var to support LD variations based on dev email

vk-outreach:
**What this PR does / why we need it**:   Plumbs LD_DEV_EMAIL
**JIRA ID**: DT-0
**Notes for your reviewer**:

The flagship client supports an env var LD_DEV_EMAIL.  When this is specified, all Launch Darkly requests include this as a custom flag value.  This allows feature flags to be changed for specific dev env hosts.

This PR adds a similar setup (but using DEV_EMAIL and dev_email so that this can be used in more places than just LD).  Also the `DEV_EMAIL` env var is just an override with the default being `git config user.email`

The plan is to make the following changes (all in separate repos sadly) to support this:

1. Update the bootstrapped fflags.yaml jsonnet to use `std.extVar('dev_email')` so that the env var gets picked up by bootstrapped services automatically.
2. Update flagship for similar support (likely by updating the config map + plumbing through the flagship to the runtime_flags library)
~3. Update `orc setup` to modify the default shellrc to setup this env var based on the employee email in okta~

Basically a series of one-line changes to plumb this through everywhere.  Feel free to suggest an alternate solution.

### DIFF
--- a/shell/deploy-to-dev.sh
+++ b/shell/deploy-to-dev.sh
@@ -60,5 +60,5 @@ kubecfg \
   -V environment="$environment" \
   -V version="$version" \
   -V bento="$bento" \
-  -V ld_dev_email="$LD_DEV_EMAIL" \
+  -V dev_email="${DEV_EMAIL:-$(git config user.email)}" \
   -V host="bento1a.outreach-dev.com"
--- a/shell/deploy-to-dev.sh
+++ b/shell/deploy-to-dev.sh
@@ -60,4 +60,5 @@ kubecfg \
   -V environment="$environment" \
   -V version="$version" \
   -V bento="$bento" \
+  -V ld_dev_email="$LD_DEV_EMAIL" \
   -V host="bento1a.outreach-dev.com"


deregtd (approved):

deregtd:
seems like a fine approach to me

richardlarocque:
I really like this.  I like it so much I'd prefer to make it not LD-specific.  I'd like to have something like this in dev-env traces so we can provide links to pre-cooked Honeycomb views that show devs their own environment and no one else's.

So my first request would be to drop the LD prefix.  Can we call this just `DEV_EMAIL`?

Also, when it comes to setting this value, maybe we can avoid the bash ENV var fiddling.  I think `git config user.email` would be a good default, and we can let folks override it with an explicit ENV var if they want to.

vk-outreach:
The client jsx code already has references to LD_DEV_EMAIL and the feature flag in LD is currently named ld_dev_email.

But I like the idea of plumbing this in DD & traces too, so will make a clean start using DEV_EMAIL and dev_email.  I plan to update giraffe as well.

I like the idea of using `git config user.email` with an override from DEV_EMAIL if it is set.  The `git config user.email` also requires no changes to orc, which is nice.

jeff-outreach:
> The client jsx code already has references to LD_DEV_EMAIL and the feature flag in LD is currently named ld_dev_email.
> 
> But I like the idea of plumbing this in DD & traces too, so will make a clean start using DEV_EMAIL and dev_email. I plan to update giraffe as well.
> 
> I like the idea of using `git config user.email` with an override from DEV_EMAIL if it is set. The `git config user.email` also requires no changes to orc, which is nice.

I concur with using `git config user.email`.

vk-outreach:
Updated to use `git config user.email` unless `DEV_EMAIL` is specified.   I can't think of any issues with invoking git config from within this script -- hopefully this isn't invoked in a container in some code path.

deregtd (approved):

PR #7, opened 2021-05-10
fix: ruby client app name fix

kadukf-outreach:
It was probably copy/paste issue but appName for ruby client was hardcoded. Following same patter as in other sh files.

**JIRA ID**: CMS-637

**Notes for your reviewer**:


jaredallard (approved):
Thanks!

PR #8, opened 2021-05-10
fix: return proper resource version

grevych:
Resource versions query in bootstrap lib currently returns `null` for non-existent resources which leads test script to setup non-existent containers. This PR adds a fix to the mentioned query.


jaredallard (approved):
Good catch. Thanks!

PR #10, opened 2021-05-13
feat: add debug script from bootstrap

malept:
**What this PR does / why we need it**:  This adds a debug script that was omitted when we moved the repository scripts to this repo. It also adds a helpful `make debug` command which works similarly to `make devserver`.

**JIRA ID**: DT-00

**Notes for your reviewer**: I did a quick smoke test on an example repository to make sure that delve actually ran. (The old, deleted version did not work with the recent gobin script).


jaredallard (approved):
LGTM. Thanks for adding this!

PR #15, opened 2021-06-25
feat: add pre-action steps and use double-colon to Makefile

vk-outreach:
**What this PR does / why we need it**:   Add pre and post action support to Makefile
**JIRA ID**: DWF-5884
**Notes for your reviewer**:

This adds support for individual repos to add pre and post actions to targets that were meaningful and not internal.

The pre action is via a `pre-xyz` target.  The post action is just using the [double-colon](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html) support in make.

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -50,7 +50,6 @@ pre-debug::
 pre-docker-build::
 pre-fmt::
 
-
 ## release:         tag a new release with goreleaser
 .PHONY: release
 release:: pre-release
--- a/root/Makefile
+++ b/root/Makefile
@@ -160,12 +160,12 @@ fmt:: pre-fmt
 
 ## fly-login:       log into concourse
 .PHONY: fly-login
-fly-login::
+fly-login:
 	${FLY} -t devs userinfo || ${FLY} -t devs login --concourse-url https://concourse.outreach.cloud -n devs
 
 ## update-pipeline: update application pipeline in concourse
 .PHONY: update-pipeline
-update-pipeline:: fly-login
+update-pipeline: fly-login
 	@echo " ===> updating concourse pipeline <==="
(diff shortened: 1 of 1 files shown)


root/Makefile
```diff
@@ -35,14 +35,30 @@ BASE_TEST_ENV       ?= GOPROXY=$(GOPROXY) GOPRIVATE=$(GOPRIVATE) OUTREACH_ACCOUN
...
+pre-integration::
+pre-e2e::
+pre-benchmark::
+pre-gogenerate::
+pre-devserver::
+pre-debug::
+pre-docker-build::
+pre-fmt::
+
+
```
malept, line 53: Extra newline?

root/Makefile
```diff
@@ -35,14 +35,30 @@ BASE_TEST_ENV       ?= GOPROXY=$(GOPROXY) GOPRIVATE=$(GOPRIVATE) OUTREACH_ACCOUN
...
 ## release:         tag a new release with goreleaser
 .PHONY: release
-release:
+release:: pre-release
 	@# Create a tag for our version
 	@git tag -d "$(APP_VERSION)" >&2 || true
 	@git tag "$(APP_VERSION)" >&2
 	@./.bootstrap/shell/gobin.sh github.com/goreleaser/goreleaser@v0.157.0 release --skip-publish --rm-dist
-	@# Delete the tag once we're done.
+	@# Delete the tag once we\'re done.
```
malept, line 61: That really needs to be escaped? If so I'd rather just avoid it entirely.
```suggestion
	@# Delete the tag once we are done.
```

malept (approved):
Looks legit. I've never used the double colon targets before.

My comments are non-blocking.

deregtd (approved):

PR #17, opened 2021-07-06
feat: add back twistlock scanning support

jaredallard:
**What this PR does / why we need it**: This PR adds back support for scanning Docker images via twistlock, and brings perf improvements for building docker images on PRs. 

**JIRA ID**: DT-0

**Notes for your reviewer**:

PR #18, opened 2021-07-08
fix(prismaci): copy image scanner from prismaci repo into devbase and reenable it

nissim-natanov-outreach:
This PR is the first to support [DT-509]: re-enable image scan in bootstrap and move twist-scan shell from prismaci repo into devbase to run in machine-mode on a bare VM itself. The original script was slightly modified to support twistcli download shortly before running the scan (and I removed the curl -k flag that could cause security issues - it should not be needed).

**JIRA ID**: [DT-509]

**Notes for your reviewer**:
This PR has been tested by faking bootstrap: devenv to point to private branch + temporary removing CIRCLE_TAG block to enable scan on all runs (and not just release/tags).

[DT-509]: https://outreach-io.atlassian.net/browse/DT-509
[DT-509]: https://outreach-io.atlassian.net/browse/DT-509

### DIFF
--- a/shell/docker-builder.sh
+++ b/shell/docker-builder.sh
@@ -28,13 +28,14 @@ args=("--ssh" "default" "--progress=plain" "--file" "deployments/${appName}/Dock
 
 # Build a quick native image on PRs and load it into docker cache
 # for security scanning
-if [[ -z $CIRCLE_TAG ]]; then
+# TODO: before final merge: uncomment if/fi
+#if [[ -z $CIRCLE_TAG ]]; then
   info "Building Docker Image (test)"
   docker buildx build "${args[@]}" -t "${appName}" --load .
 
   info "🔐 Scanning docker image for vulnerabilities"
   source "${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}"
-fi
+#fi
 
 if [[ -n ${CIRCLE_TAG} ]]; then
   echo "🔨 Building and Pushing Docker Image (production)"
--- a/shell/docker-builder.sh
+++ b/shell/docker-builder.sh
@@ -36,7 +36,7 @@ if [[ -z $CIRCLE_TAG ]]; then
   source "${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}"
 fi
 
-if [[ -n ${CIRCLE_TAG} ]]; then
+if [[ -n $CIRCLE_TAG ]]; then
   echo "🔨 Building and Pushing Docker Image (production)"
   set -x
   docker buildx build "${args[@]}" --platform linux/arm64,linux/amd64 \
--- a/shell/docker-builder.sh
+++ b/shell/docker-builder.sh
@@ -44,6 +44,6 @@ set +x
 # TODO: Re-enable this when buildx supports exporting multiple-manifest docker images
 #if [[ -z $CIRCLE_TAG ]]; then
 #  # Scan the built image
-#  info "Scanning docker image for vulnerabilities"
-#  /usr/local/bin/twist-scan.sh "$appName"
+  info "Scanning docker image for vulnerabilities"
+  /usr/local/bin/twist-scan.sh "$appName"
 #fi
--- a/shell/docker-builder.sh
+++ b/shell/docker-builder.sh
@@ -28,14 +28,13 @@ args=("--ssh" "default" "--progress=plain" "--file" "deployments/${appName}/Dock
 
 # Build a quick native image on PRs and load it into docker cache
 # for security scanning
-# TODO: before final merge: uncomment if/fi
-#if [[ -z $CIRCLE_TAG ]]; then
+if [[ -z $CIRCLE_TAG ]]; then
   info "Building Docker Image (test)"
   docker buildx build "${args[@]}" -t "${appName}" --load .
 
   info "🔐 Scanning docker image for vulnerabilities"
   source "${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}"
-#fi
+fi
 
 if [[ -n ${CIRCLE_TAG} ]]; then
   echo "🔨 Building and Pushing Docker Image (production)"
--- a/shell/docker-builder.sh
+++ b/shell/docker-builder.sh
@@ -24,27 +24,23 @@ source "${LIB_DIR}/buildx.sh"
 # shellcheck source=./lib/ssh-auth.sh
 source "${LIB_DIR}/ssh-auth.sh"
 
-extraArgs=("-t" "$remote_image_name:${VERSION}")
-if [[ -n ${CIRCLE_TAG} ]]; then
-  # Only push on a tag
-  extraArgs+=("--push")
-else
-  echo "Note: Skipping twist-scan due to buildx limitations"
-  # TODO: buildx doesn't currently support this.
-  # Load it into the docker cache so we can run twist-scan
-  #extraArgs+=("--load")
-fi
+args=("--ssh" "default" "--progress=plain" "--file" "deployments/${appName}/Dockerfile" "--build-arg" "VERSION=${VERSION}")
 
-echo "🔨 Building Docker Image"
-set -x
-docker buildx build --ssh default --progress=plain \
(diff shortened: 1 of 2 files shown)


jaredallard:
This can't be re-enabled without https://github.com/getoutreach/devbase/pull/17 otherwise it will fail due to the docker image not existing in the host docker daemon. I recommend reading up on [buildkit](https://docs.docker.com/buildx/working-with-buildx/).

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+  exit 1
+fi
+if ! [ -x "$(which curl)" ]; then 
+  echo "We need an executable called curl in the $PATH in order to execute"
+  exit 1
+fi
+if ! [ -x "$(which jq)" ]; then 
+  echo "We need an executable called curl in the $PATH in order to execute"
+  exit 1
+fi
```
jaredallard, line 20: style: prefer built-ins to external programs
```suggestion
if ! command -v curl >/dev/null 2>&1; then 
  echo "We need an executable called curl in the $PATH in order to execute"
  exit 1
fi
if ! command -v jq >/dev/null 2>&1; then 
  echo "We need an executable called curl in the $PATH in order to execute"
  exit 1
fi
```

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+# This script scans docker images for vulnerabilities, generates local
+# report and also uploads the report to Prisma Cloud. The script expects
+# special credentials - those are set by prismacloud-credentials circleci context.
+#
+
+## Add some safeties for external dependencies when running
+if [ -z "${CIRCLECI}" ]; then
+  echo "I can only run in CircleCI"
+  exit 1
+fi
```
jaredallard, line 12: style: Use double brackets where possible, also `${}` is not really needed in general, try to use only when needed
```suggestion
if [[ -z $CIRCLECI ]]; then
  echo "I can only run in CircleCI"
  exit 1
fi
```

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+if [ -z "${PC_ACCESS_KEY}" ]; then
+  echo "Need an environment variable called PC_ACCESS_KEY"
+  exit 1
+fi
+if [ -z "${PC_SECRET_KEY}" ]; then
+  echo "Need an environment variable called PC_SECRET_KEY"
+  exit 1
+fi
```
jaredallard, line 32: style: Same goes for these, double brackets and no `{}`

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+PC_CLOUD_TOKEN=$(jq -r -c '.token' <${PC_CLOUD_TOKEN_JSON})
+
+PC_CLOUD_TWISTCLI=/tmp/twistcli
+info "downloading twistcli from prismacloud"
+curl -L --header "authorization: Bearer ${PC_CLOUD_TOKEN}" \
+  "${PC_CONSOLE_URL}/api/v1/util/twistcli" -o ${PC_CLOUD_TWISTCLI}
+chmod a+x ${PC_CLOUD_TWISTCLI}
+
+
+if [ -z "${DOCKER_HOST}" ]; then
```
jaredallard, line 52:
```suggestion
if [[ -z $DOCKER_HOST ]]; then
```

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+    # trying the default unix one
+    TWISTCLI_DOCKER_HOST=/var/run/docker.sock
+else
+    TWISTCLI_DOCKER_HOST=$(echo "${DOCKER_HOST}" | sed -e 's/tcp/https/')
```
jaredallard, line 56: nit: we don't need an echo here

```suggestion
    TWISTCLI_DOCKER_HOST=$(sed -e 's/tcp/https/' <<< "${DOCKER_HOST}")
```

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
```
jaredallard, line 1: Use bash if possible

```suggestion
#!/usr/bin/env bash
```

shell/security/prismaci/twist-scan.sh
nissim-natanov-outreach: done

shell/security/prismaci/twist-scan.sh
```diff
@@ -0,0 +1,65 @@
...
+else
+    TWISTCLI_DOCKER_HOST=$(sed -e 's/tcp/https/' <<< "${DOCKER_HOST}")
+fi
+
+${PC_CLOUD_TWISTCLI} images scan --token "${PC_CLOUD_TOKEN}" \
+  --ci --details --address "${PC_CONSOLE_URL}" \
+  --custom-labels \
+  --docker-address "${TWISTCLI_DOCKER_HOST}" \
+  --output-file "${TEST_RESULTS}/image_scan.json" \
+  "$@"
```
edyesed, line 64: This would not work in this form inside the docker executor in circleci, I am not sure of the behavior in the machine executor.

There are env variables that docker client uses by convention in order to identify DOCKER_HOST and certificate materials for the docker client to use in order to validate a TLS'ed connection via dockerd.  If DOCKER_HOST is not the empty string, the twistcli is going to need the  `--docker-tlscert`, `--docker-tlscacert`, and `--docker-tlskey` flags  to be set.

nissim-natanov-outreach: Bootstrap is going to run this script on a bare VM, not inside docker. Since docker address is going to be pure unix sock, TLS certs are no longer needed. BTW - this PR has been fully tested on a bootstraptestservice - and it prodused scan results :).

edyesed: great, then I would suggest either removing the test for DOCKER_HOST, or making that test fail if DOCKER_HOST is not the empty string. As is, this form of the twistcli command would be non-functional if DOCKER_HOST is not the empty string.

shell/docker-builder.sh
```diff
@@ -3,47 +3,43 @@
...
-#if [[ -z $CIRCLE_TAG ]]; then
-#  # Scan the built image
-#  info "Scanning docker image for vulnerabilities"
-#  /usr/local/bin/twist-scan.sh "$appName"
-#fi
+if [[ -n $CIRCLE_TAG ]]; then
+  echo "🔨 Building and Pushing Docker Image (production)"
+  set -x
+  docker buildx build "${args[@]}" --platform linux/arm64,linux/amd64 \
+    -t "${remote_image_name}:${VERSION}" --push .
```
jaredallard, line 43: While we are here, we can fix the bug just reported in dev-tooling-support.

```suggestion
    -t "${remote_image_name}:${VERSION}" -t "$remote_image_name:$VERSION" --push .
```

shell/security/prismaci/twist-scan.sh
nissim-natanov-outreach: updated

jaredallard (approved):
LGTM

shell/docker-builder.sh
```diff
@@ -3,47 +3,43 @@
...
-#if [[ -z $CIRCLE_TAG ]]; then
-#  # Scan the built image
-#  info "Scanning docker image for vulnerabilities"
-#  /usr/local/bin/twist-scan.sh "$appName"
-#fi
+if [[ -n $CIRCLE_TAG ]]; then
+  echo "🔨 Building and Pushing Docker Image (production)"
+  set -x
+  docker buildx build "${args[@]}" --platform linux/arm64,linux/amd64 \
+    -t "${remote_image_name}:${VERSION}" -t "$remote_image_name:latest" --push .
```
edyesed, line 43: `-t "$remote_image_name:latest"` might be more consistently expressed as `-t "${remote_image_name}:latest"`

PR #25, opened 2021-07-21
feat(protoc):Updating to a more recent version of grpc-js

masterkidan:
This PR updates devbase to use the latest grpc tooling. The goal here is to allow generation of grpc client libs that use grpc-js instead of the the now deprecated 'grpc' lib.

**JIRA ID**: [DT-355]

**Notes for your reviewer**:

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -47,8 +47,7 @@ docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l go 
 if has_grpc_client "node"; then
   info_sub "node"
   docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l node \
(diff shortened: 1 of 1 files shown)


shell/protoc.sh
```diff
@@ -47,7 +47,8 @@ docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l go
 if has_grpc_client "node"; then
   info_sub "node"
   docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l node \
-    --with-typescript -o "./clients/node/src/grpc/"
+    --with-typescript -o "./clients/node/src/grpc/" --grpc-out="service=grpc-node,mode=grpc-js:$OUT_DIR" \
```
masterkidan, line 50: Please ignore these changes

PR #26, opened 2021-07-26
feat: skip e2e based off build tags

george-e-shaw-iv:
* Conditionally skips e2e tests if no `or_e2e` build tags are found.

**JIRA ID**: XX-XX

**Notes for your reviewer**:

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -2,9 +2,14 @@ package main
 
 import (
 	"context"
+	"go/ast"
+	"go/parser"
+	"go/token"
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
+	"strings"
 	"time"
 
 	dockerclient "github.com/docker/docker/client"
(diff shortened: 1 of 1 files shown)


e2e/e2e.go
```diff
@@ -192,6 +197,70 @@ func main() {
...
+			// Skip symlinks.
+			return nil
+		}
+
+		if !strings.HasSuffix(path, "_test.go") {
+			// Skip all files that aren't go test files.
+			return nil
+		}
+
+		f, err := os.ReadFile(path)
```
george-e-shaw-iv, line 225: lol guess who forgot to close the file again

george-e-shaw-iv: wait i don't have to in this case. played myself.

jaredallard: yeah I was about to comment, I don't think reading it into memory is a great idea if we can avoid it

jaredallard (approved):
This is a great QoL improvement, thanks! We can address `os.ReadFile` now, or later, not a huge deal

e2e/e2e.go
```diff
@@ -192,6 +197,70 @@ func main() {
...
+			// No need to keep walking through files if we've already found one file
+			// that requries e2e tests.
+			return nil
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
plisy, line 215: Is this because we are still on go 1.15 and WalkDir was introduced in 1.16?

plisy (approved):

e2e/e2e.go
george-e-shaw-iv: yeah we're not completely switched over to 1.16 as a whole so i think it's safe to continue to use filepath.Walk

e2e/e2e.go
```diff
@@ -192,6 +197,78 @@ func main() {
 
 	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
 
+	// runEndToEndTests is a flag that denotes whether or not this needs to actually
+	// run or not based off of the filepath.Walk function immediately proceeding.
+	var runEndToEndTests bool
+
+	err := filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
```
vk-outreach, line 204: I think it might be simpler to use [build.ImportDir](https://pkg.go.dev/go/build#Context.ImportDir) and inspect the [Package.AllTags](https://pkg.go.dev/go/build#Package).   If you decide to just keep the current structure, I suggest atleast using [constraint](https://pkg.go.dev/go/build/constraint@go1.16.6) to check if a line is a build constraint.

george-e-shaw-iv: Oh I didn't see this before hitting the merge button. I'll change the code to use the `build.ImportDir` function whenever I get some free time next, that seems like a better way to do this, I agree. I'll create a ticket to track this

george-e-shaw-iv: https://outreach-io.atlassian.net/browse/DTSS-605

PR #32, opened 2021-08-02
feat(fmt): run eslint --fix on the Node.js API client

malept:
**What this PR does / why we need it**: This fixes a formatting bug where `eslint` may have formatting changes (such as import order) but `make fmt` doesn't apply them, which means that rebootstrapping reverts the changes.

See, for example, https://github.com/getoutreach/authz/pull/465

**JIRA ID**: DTSS-00

**Notes for your reviewer**:


jaredallard (approved):

malept:
FWIW I did test this change on authz and it worked as expected.

PR #35, opened 2021-08-04
fix: enable any host to connect to the local mysql container

kadukf-outreach:
**What this PR does / why we need it**: it is not possible to connect to the mysql server over cli when the mysql container is started. The change enables it. 

**JIRA ID**: CMS-704

**Notes for your reviewer**:

### DIFF
--- a/shell/test.sh
+++ b/shell/test.sh
@@ -69,7 +69,7 @@ if [[ -z $CI && $TEST_TAGS == *"or_int"* ]]; then
 
   if has_resource "mysql"; then
     info_sub "creating mysql container"
-    mysqlID=$(docker run -itd --rm -p 3306:3306 -e "MYSQL_DATABASE=$(get_app_name)" -e MYSQL_ROOT_PASSWORD=root "gcr.io/outreach-docker/mysql:$(get_resource_version "mysql")")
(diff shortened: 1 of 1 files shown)


shell/test.sh
```diff
@@ -69,7 +69,7 @@ if [[ -z $CI && $TEST_TAGS == *"or_int"* ]]; then
 
   if has_resource "mysql"; then
     info_sub "creating mysql container"
-    mysqlID=$(docker run -itd --rm -p 3306:3306 -e "MYSQL_DATABASE=$(get_app_name)" -e MYSQL_ROOT_PASSWORD=root "gcr.io/outreach-docker/mysql:$(get_resource_version "mysql")")
+    mysqlID=$(docker run -itd --rm -p 3306:3306 -e "MYSQL_DATABASE=$(get_app_name)" -e "MYSQL_ROOT_HOST=%" -e MYSQL_ROOT_PASSWORD=root "gcr.io/outreach-docker/mysql:$(get_resource_version "mysql")")
```
masterkidan, line 72: This would only expose the docker container within the k8 cluster right? Can you clarify how this allows external 'hosts' to connect... 
Don't you also need an ingress/nodeport or something configured in k8 for this to work?

masterkidan: Oh I see, this is only for the e2e tests. Seems reasonable to me, @jaredallard  for thoughts?

jaredallard:
Closing due to age, had a conversation on the other PR. Can chat offline if still needed.

PR #36, opened 2021-08-05
feat: kill localizer if e2e spawns it [DTSS-595]

george-e-shaw-iv:

**JIRA ID**: [DTSS-595]

**Notes for your reviewer**:

[DTSS-595]: https://outreach-io.atlassian.net/browse/DTSS-595


jaredallard (approved):

e2e/e2e.go
```diff
@@ -299,16 +301,37 @@ func main() {
...
+		cmd = exec.CommandContext(ctx, "devenv", "--skip-update", "tunnel")
+		cmd.Stderr = os.Stderr
+		cmd.Stdout = os.Stdout
+		cmd.Stdin = os.Stdin
+		err = cmd.Start()
+		if err != nil {
+			log.Fatal().Err(err).Msg("Failed to start devenv tunnel")
+		}
+
+		tick := time.NewTicker(2 * time.Second)
```
vk-outreach, line 319: you can simplify this a lot by using [async.Sleep](https://github.com/getoutreach/async/blob/fbcac8debd5ae5fa41e887f3de7133cc23ac2244/pkg/async/async.go#L132)

```golang
for ctx.Err() != nil && !localize.IsRunning() {
     async.Sleep(ctx, time.Second*2)
}
```

e2e/e2e.go
```diff
@@ -319,4 +343,30 @@ func main() {
...
+
+		cmd = exec.CommandContext(ctx, "sudo", "kill", strings.TrimSpace(buf.String()))
+		cmd.Stderr = os.Stderr
+		cmd.Stdout = os.Stdout
+		cmd.Stdin = os.Stdin
+		err = cmd.Start()
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to kill localizer using process ID")
+		}
+	}
```
vk-outreach, line 371: We should probably move this whole thing to some helper function rather than inline.  I doubt this is the last time we want to kill a proc by name.

e2e/e2e.go
```diff
@@ -299,16 +302,37 @@ func main() {
...
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
-	cmd.Stdin = os.Stdin
-	err = cmd.Start()
-	if err != nil {
-		log.Fatal().Err(err).Msg("Failed to start devenv tunnel")
+	var killLocalizer bool
+	if !localizer.IsRunning() {
+		killLocalizer = true
+
```
vk-outreach, line 308: My last comment here seems to have been dropped on the floor by github :(

Anyway, I think we should use the standard defer/cleanup pattern here (maybe using [cleanup](https://engdocs.outreach.cloud/github.com/getoutreach/gobox/pkg/cleanup):

```suggestion
	var killLocalizer func()
	cleanups := cleanup.Funcs{&killLocalizer} // can add more cleanups later on if we need to
	defer cleanups.Run()
	
	if !localizer.IsRunning() {
		killLocalizer = func() { killProcByName("localizer") }
		... start localizer... 
	}
```

vk-outreach:
(none of my comments are blocking, ofcourse)

e2e/e2e.go
george-e-shaw-iv: Yeah I'll do this, sounds reasonable

e2e/e2e.go
george-e-shaw-iv: So I ended up having to go a completely different direction because it was proving to be impossible to get the PID of that process using shell commands. I ended up implementing a kill RPC in the localizer server. This seemed like overall a better option for not only this, but also for devenv tunnel. We'll have more control to clean up users potentially messed up environments.

e2e/e2e.go
george-e-shaw-iv: there ended up being not a reason to do this quite yet, really not much to cleanup other than localizer

jaredallard (approved):

PR #37, opened 2021-08-05
fix: update buildx

asms:
Update buildx so that docker build can accept secrets from environment
variables.

**JIRA ID**: XX-XX

**Notes for your reviewer**:


jaredallard (approved):
Didn't realize this was that new, whew. :shipit:

PR #38, opened 2021-08-12
fix: remove merging of configuration

george-e-shaw-iv:

This PR moves the merging of configuration functionality in local development. Instead, users should be modifying the `<appName>.config.jsonnet` file instead, which is more deterministic and compatible with the Tilt world we'll be going to soon. Users should use `make devserver` to ensure their changes are always reflected locally.

[**CONTEXT**](https://outreach-hq.slack.com/archives/CN9MU7GLW/p1628786338329700)

**JIRA ID**: XX-XX

**Notes for your reviewer**:

### DIFF
--- a/shell/devconfig.sh
+++ b/shell/devconfig.sh
@@ -88,17 +88,9 @@ DEPLOY_TO_DEV_ENVIRONMENT=local_development "$DIR/deploy-to-dev.sh" show | yq -r
     echo -e "$configData" | sed "s:/run/secrets/outreach.io:$configDir:g" |
       sed "s:/run/volumes/outreach.io:$volumeDir:g" >"$tmpFile"
 
-    # If the file already exists, then merge it with the new one.
-    # NOTE: We probably want to make this smarter and log when merging. I don't think
-    # that is feasible to do in bash, though....
-    if [[ -e $saveFile ]] && [[ ${saveFile##*.} =~ ^(yaml|yml)$ ]]; then
-      info_sub "$configFile (merged)"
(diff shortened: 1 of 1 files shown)


shell/devconfig.sh
```diff
@@ -88,17 +88,9 @@ DEPLOY_TO_DEV_ENVIRONMENT=local_development "$DIR/deploy-to-dev.sh" show | yq -r
...
-    if [[ -e $saveFile ]] && [[ ${saveFile##*.} =~ ^(yaml|yml)$ ]]; then
-      info_sub "$configFile (merged)"
-      yq --yaml-output -s '.[0] * .[1]' "$tmpFile" "$saveFile" >"$mergedFile"
-      mv "$mergedFile" "$saveFile"
-    else
-      info_sub "$configFile"
-      mv "$tmpFile" "$saveFile"
-    fi
+    info_sub "$configFile"
+    mv "$tmpFile" "$saveFile"
```
jaredallard, line 92: As mentioned in our internal slack, this was a feature requested by people to allow local modifications to that configuration to be persisted across `make devserver` runs. Not sure how to fix this while retaining that functionality at the moment

masterkidan: As a general qq: In this particular case can we not make this a flag that we pass in to the devenv command? that way we can support both... 

It feels like most users would not care for this merging. Does that seem reasonable? 
Something like --persist-config ?

jaredallard: That's a question for @deregtd as the main person wanting this feature, I think.

jaredallard (approved):

PR #41, opened 2021-08-20
feat: support junit output for unit/e2e tests

jaredallard:
**What this PR does / why we need it**: Title says, it does. 

**JIRA ID**: [DTSS-0]

**Notes for your reviewer**:

### DIFF
--- a/shell/test.sh
+++ b/shell/test.sh
@@ -128,6 +128,11 @@ set -ex
 # Why: We want these to split. For those wondering about "$@":
 # https://stackoverflow.com/questions/5720194/how-do-i-pass-on-script-arguments-that-contain-quotes-spaces
 # shellcheck disable=SC2086
-go test $BENCH_FLAGS $COVER_FLAGS $TEST_FLAGS \
+"$DIR/gobin.sh" gotest.tools/gotestsum@v1.7.0 --junitfile bin/unit-tests.xml -- $BENCH_FLAGS $COVER_FLAGS $TEST_FLAGS \
   -ldflags "-X github.com/getoutreach/go-outreach/v2/pkg/app.Version=testing -X github.com/getoutreach/gobox/pkg/app.Version=testing" -tags="$TEST_TAGS" \
   "$@" ./...
+
+if [[ -n $CI ]]; then
+  mkdir -p /tmp/test-results
+  mv bin/unit-tests.xml /tmp/test-results
+fi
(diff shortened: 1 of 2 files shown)


george-e-shaw-iv (changes requested):
adds too much code, try and simplify this please

shell/test.sh
```diff
@@ -128,6 +128,11 @@ set -ex
 # Why: We want these to split. For those wondering about "$@":
 # https://stackoverflow.com/questions/5720194/how-do-i-pass-on-script-arguments-that-contain-quotes-spaces
 # shellcheck disable=SC2086
-go test $BENCH_FLAGS $COVER_FLAGS $TEST_FLAGS \
+"$DIR/gobin.sh" gotest.tools/gotestsum@v1.7.0 --junitfile bin/unit-tests.xml -- $BENCH_FLAGS $COVER_FLAGS $TEST_FLAGS \
   -ldflags "-X github.com/getoutreach/go-outreach/v2/pkg/app.Version=testing -X github.com/getoutreach/gobox/pkg/app.Version=testing" -tags="$TEST_TAGS" \
   "$@" ./...
+
+if [[ -n $CI ]]; then
```
masterkidan, line 135: just for understanding: can you add a comment here as to why this move in file directory is required? Guessing it has to do with circle ci , but would be good to explicitly call it out.

shell/test.sh
```diff
@@ -128,6 +128,11 @@ set -ex
 # Why: We want these to split. For those wondering about "$@":
 # https://stackoverflow.com/questions/5720194/how-do-i-pass-on-script-arguments-that-contain-quotes-spaces
 # shellcheck disable=SC2086
-go test $BENCH_FLAGS $COVER_FLAGS $TEST_FLAGS \
+"$DIR/gobin.sh" gotest.tools/gotestsum@v1.7.0 --junitfile bin/unit-tests.xml -- $BENCH_FLAGS $COVER_FLAGS $TEST_FLAGS \
```
masterkidan, line 131: Rather than lock it to 1.7.0 explicitly mind putting this in an environment variable? Just makes it more configurable

masterkidan (approved):

shell/test.sh
vk-outreach: We don't really use environment variables (and we don't want to start doing that either).  

I think we should use the service.yaml versions override setup for this like we do for other tools.

jaredallard: Fixed

shell/test.sh
jaredallard: Fixed

george-e-shaw-iv (approved):

PR #43, opened 2021-08-27
feat(OLM-952): support importing proto files from schema repo

zaquestion:

These changes allow settings service to move forward with leveraging
settings payloads defined in the `schema` repo directly in the settings
service.

**JIRA ID**: [OLM-952]

**Notes for your reviewer**:

Also of note: This only account for generating the pb code within
the repo itself. It won't generate the imported code, that step needs to
be done manually for the service and it's clients

Generation tested manually with and without imports

Example:
```
go-protoc-imports:
  - "github.com/org/repo@latest"
```

[OLM-952]: https://outreach-io.atlassian.net/browse/OLM-952

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -58,7 +58,6 @@ docker exec "$CONTAINER_ID" sh -c "groupadd -f --gid $gid localuser && useradd -
 info_sub "go"
 docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l go \
   $(for import in $(get_list "go-protoc-imports"); do echo "-i /mod/$(get_import_basename "$import")"; done) \
-  $(if has_feature "validation"; then echo "--with-validator --validator-source-relative"; fi) \
   --go-source-relative -o ./
 
 if has_grpc_client "node"; then
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -29,9 +29,24 @@ if [[ -z $uid ]] || [[ -z $gid ]]; then
   gid=1000
 fi
 
+get_import_basename() {
+  basename $(go list -f '{{ .Path }}' -m "$1" | sed 's|/v[0-9]||') # sed removes the version off the module path if present
+}
+for import in $(get_list "go-protoc-imports"); do
+  go install "$import"
+  # check exit for this instead of the install itself, as the install can
+  # return non-zero, but still retrieve the dependency. As long as we can
+  # retrieve the dep with go list, we can import the files for protoc
+  go list -f '{{ .Dir }}' -m "$import"
+  exit_code=$(echo $?)
+  if [[ $exit_code != "0" ]]; then
+    exit $exit_code
+  fi
+done
 # Create the protoc container.
 info "Generating GRPC Clients"
 CONTAINER_ID=$(docker run --rm -v "$(get_repo_directory)/api:/defs" \
+  $(for import in $(get_list "go-protoc-imports"); do echo "-v $(go list -f '{{ .Dir }}' -m $import):/mod/$(get_import_basename "$import")"; done) \
   --entrypoint bash -d "$IMAGE" -c 'exec tail -f /dev/null')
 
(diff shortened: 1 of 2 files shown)


zaquestion:
Changed this to support a more generic means of importing proto files via go modules.

shell/protoc.sh
```diff
@@ -29,9 +29,24 @@ if [[ -z $uid ]] || [[ -z $gid ]]; then
   gid=1000
 fi
 
+get_import_basename() {
+  basename $(go list -f '{{ .Path }}' -m "$1" | sed 's|/v[0-9]||') # sed removes the version off the module path if present
+}
+for import in $(get_list "go-protoc-imports"); do
+  go install "$import"
```
jaredallard, line 36: Repeating an earlier comment, can we use `gobin` here? Downside is we'd need to manually pass the executable path to protoc

shell/protoc.sh
```diff
@@ -42,16 +57,20 @@ docker exec "$CONTAINER_ID" sh -c "groupadd -f --gid $gid localuser && useradd -
 # Create the language specific clients
 info_sub "go"
 docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l go \
+  $(for import in $(get_list "go-protoc-imports"); do echo "-i /mod/$(get_import_basename "$import")"; done) \
+  $(if has_feature "validation"; then echo "--with-validator --validator-source-relative"; fi) \
```
jaredallard, line 61: What's this validation functionality?

zaquestion: I can break this into a separate change, but this basically enables using https://github.com/envoyproxy/protoc-gen-validate with our stuff. It's the other piece that settings service needs to get fully unblocked on managing its schema. Any opposition to bringing it in, in general?

shell/protoc.sh
zaquestion: I briefly looked at this, I might be misunderstanding `gobin`s use here. It looked like our script was for executing binaries. The purpose of this install isn't to execute anything, but rather to get the dependency pulled into the modules cache.

shell/protoc.sh
vk-outreach: Can we have a discussion on this in #go before we make this choice?  I don't have a strong opinion on it itself but if we do this we are effectively encouraging folks to use `protoc-gen-validate` as the way to validate constraints and seems like we should see if anyone else has any opinions on this.

jaredallard: Ah, no opposition just wasn't super sure what it was!

shell/protoc.sh
jaredallard: Right, but `go install` doesn't work great on <1.16, so we've used gobin to do this for the time being.

shell/protoc.sh
zaquestion: I'll go ahead and pull this out into it's own PR and get a discussion going in #go

shell/protoc.sh
zaquestion: Added a version check and tested with
```
$ go version | awk '{ print $3 }' | sed 's|go||'
1.16.6
$ /home/zaq/projects/outreach/settings/scripts/shell-wrapper.sh protoc.sh
go install github.com/getoutreach/schema@latest: module github.com/getoutreach/schema@latest found (v0.0.0-20210830182546-82ce20ed15eb), but does not contain package github.com/getoutreach/schema
/home/zaq/.asdf/installs/golang/1.16.6/packages/pkg/mod/github.com/getoutreach/schema@v0.0.0-20210830182546-82ce20ed15eb
 :: Generating GRPC Clients
^C
zaq@vm:~/projects/outreach/settings (schema_import_test)
$ vim .tool-versions
zaq@vm:~/projects/outreach/settings (schema_import_test)
$ /home/zaq/projects/outreach/settings/scripts/shell-wrapper.sh protoc.sh
Go version must be greater than 1.16.0 to use 'go-protoc-imports' feature
zaq@vm:~/projects/outreach/settings (schema_import_test)
$ go version
go version go1.15.2 linux/amd64

```

jaredallard (approved):

PR #52, opened 2021-09-22
fix(opslevel): Always scan images to ensure opslevel report is sent on releases

nissim-natanov-outreach:
**What this PR does / why we need it**: we need OpsLevel report on released images, yet we only scanned and produced them for PRs.

**JIRA ID**: [DTSS-509]

**Notes for your reviewer**:

[DTSS-509]: https://outreach-io.atlassian.net/browse/DTSS-509

### DIFF
--- a/shell/docker-builder.sh
+++ b/shell/docker-builder.sh
@@ -27,15 +27,13 @@ source "${LIB_DIR}/ssh-auth.sh"
 secrets=("--secret" "id=npmtoken,env=NPM_TOKEN")
 args=("--ssh" "default" "--progress=plain" "--file" "deployments/${appName}/Dockerfile" "--build-arg" "VERSION=${VERSION}")
 
-# Build a quick native image on PRs and load it into docker cache
-# for security scanning
-if [[ -z $CIRCLE_TAG ]]; then
-  info "Building Docker Image (test)"
-  docker buildx build "${args[@]}" "${secrets[@]}" -t "${appName}" --load .
+# Build a quick native image and load it into docker cache for security scanning
+# Scan reports for release images are also uploaded to OpsLevel (test image reports only available on PR runs as artifacts).
+info "Building Docker Image (for scanning)"
+docker buildx build "${args[@]}" "${secrets[@]}" -t "${appName}" --load .
 
-  info "🔐 Scanning docker image for vulnerabilities"
-  "${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}" || echo "Warning: Failed to scan image"
-fi
+info "🔐 Scanning docker image for vulnerabilities"
+"${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}" || echo "Warning: Failed to scan image"
 
 if [[ -n $CIRCLE_TAG ]]; then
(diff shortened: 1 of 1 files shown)


edyesed (approved):
seems fine to me

shell/docker-builder.sh
```diff
@@ -27,15 +27,13 @@ source "${LIB_DIR}/ssh-auth.sh"
...
-
-  info "🔐 Scanning docker image for vulnerabilities"
-  "${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}" || echo "Warning: Failed to scan image"
-fi
+# Build a quick native image and load it into docker cache for security scanning
+# Scan reports for release images are also uploaded to OpsLevel (test image reports only available on PR runs as artifacts).
+info "Building Docker Image (for scanning)"
+docker buildx build "${args[@]}" "${secrets[@]}" -t "${appName}" --load .
+
+info "🔐 Scanning docker image for vulnerabilities"
```
masterkidan, line 35: The dev images are pushed from this repo https://github.com/getoutreach/devenv-snapshots ... perhaps I am not understanding this clearly, but should this change not happen there?

masterkidan: Oh I see, this is for any image that we build .... What happens with helmcharts ( I dont believe those use the docker build commands) etc? I would imagine this should also apply to those right?

Additionally should this scan only happen on CI? This would mean that everytime someone does a local build/bootstrap we would end up scanning for vulnerabilities right?

shell/docker-builder.sh
```diff
@@ -27,15 +27,13 @@ source "${LIB_DIR}/ssh-auth.sh"
...
-  info "Building Docker Image (test)"
-  docker buildx build "${args[@]}" "${secrets[@]}" -t "${appName}" --load .
-
-  info "🔐 Scanning docker image for vulnerabilities"
-  "${TWIST_SCAN_DIR}/twist-scan.sh" "${appName}" || echo "Warning: Failed to scan image"
-fi
+# Build a quick native image and load it into docker cache for security scanning
+# Scan reports for release images are also uploaded to OpsLevel (test image reports only available on PR runs as artifacts).
+info "Building Docker Image (for scanning)"
+docker buildx build "${args[@]}" "${secrets[@]}" -t "${appName}" --load .
```
jaredallard, line 33: In theory, this should use the build cache for the second command, if this doesn't we will need to re-evaluate this approach. For now, LGTM.

jaredallard (approved):

shell/docker-builder.sh
jaredallard: This is only used for CI, local builds use `make docker-build` in the Dockerfile. Helm charts are different, and we don't build those.

shell/docker-builder.sh
nissim-natanov-outreach: thanks Jared!

shell/docker-builder.sh
nissim-natanov-outreach: yep, this is CI only. The only diff here is that scanning will now happen for both private PRs and for release PRs (OpsLevel will be reported on release image scans only, this has not changed)

PR #54, opened 2021-09-29
feat(validate): add validator for GitHub CODEOWNERS file

malept:

This is largely an RFC. For context, there was a minor incident where a CODEOWNERS file was accidentally modified and committed to the default branch with invalid syntax, so none of the code owners were properly notified when PRs were created. As a result, I've added [codeowners-validator](https://github.com/mszostok/codeowners-validator) to the list of linters.

**JIRA ID**: DTSS-00

**Notes for your reviewer**:

I've added rudimentary support for GitHub token checking. Also, there are some hardcoded org name assumptions which need to be addressed, for which I could use some guidance.

### DIFF
--- a/shell/codeowners-validator.sh
+++ b/shell/codeowners-validator.sh
@@ -12,17 +12,17 @@ source "$DIR/lib/bootstrap.sh"
 
 CODEOWNERS_CHECKS="syntax,files,duppatterns,owners"
 if [[ -n $OUTREACH_GITHUB_TOKEN ]]; then
-  GH_TOKEN="$OUTREACH_GITHUB_TOKEN"
+	GH_TOKEN="$OUTREACH_GITHUB_TOKEN"
 elif [[ -f "$HOME/.outreach/github.token" ]]; then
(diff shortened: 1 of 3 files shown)


shell/codeowners-validator.sh
```diff
@@ -0,0 +1,28 @@
...
+	GH_TOKEN="$OUTREACH_GITHUB_TOKEN"
+elif [[ -f "$HOME/.outreach/github.token" ]]; then
+	GH_TOKEN="$(cat "$HOME/.outreach/github.token")"
+else
+	warn "No GitHub token found, skipping owner checks"
+	CODEOWNERS_CHECKS="syntax,files,duppatterns"
+fi
+
+# TODO: use org value instead of hardcoding
+OWNER_CHECKER_REPOSITORY="getoutreach/$(get_app_name)" \
```
jaredallard, line 24: Agh, I have a PR that adds support for this but I never got it merged.

jaredallard (approved):
This is dope!

PR #60, opened 2021-10-10
fix(Makefile): set default value for OSS

jaredallard:
**What this PR does / why we need it**: This fixes an issue where `OSS` would recursively call itself. With the usage of `?=` the value in `Makefile` will only be applied if not already set, so the attempt to default `OSS` that was being done before is not needed.

**JIRA ID**: [DTSS-0]

**Notes for your reviewer**:


george-e-shaw-iv (approved):
_wow_

PR #80, opened 2021-11-24
fix: ci machine perms

george-e-shaw-iv:
* permission denied errors since i was executing commands that affected directories outside of the home directory

**JIRA ID**: XX-XX

**Notes for your reviewer**:


plisy (approved):
lgtm

PR #81, opened 2021-11-29
fix(sync-package-hjson): update relative path due to the script moving

malept:

The Dependabot updater failed silently because the script moved from `shell/circleci` to `shell/ci/testing`.

## Jira ID

N/A

## Notes for your reviewers

Also added some debug info to the log for easier troubleshooting in the future.


jaredallard (approved):

jeff-outreach (approved):
Thanks for finding and fixing this @malept

PR #82, opened 2021-11-30
feat: revamp ruby, github packages, support asdf

jaredallard:
**What this PR does / why we need it**: This PR removes the hacky `run-docker-container` scripts and instead uses raw Ruby. This is now facilitated but the new multi-language docker containers powered by asdf. As part of this we also now support publishing to Github Packages instead of packagecloud

**JIRA ID**: [DTSS-0]

**Notes for your reviewer**:


shell/ci/auth/gcr.sh
```diff
@@ -2,6 +2,11 @@
 # Configures CircleCI docker authentication
```
george-e-shaw-iv, line 2: would be nice to no-op in this file if its running locally since docker login fails when running this locally via script-wrapper

shell/ci/auth/packagecloud.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# Sets up authentication for pushing up a RubyGem
+# DEPRECATED: This is no longer supported and will be replaced
+# by Github Packages very soon.
```
george-e-shaw-iv, line 4: should you log this

george-e-shaw-iv (approved):

shell/ci/auth/packagecloud.sh
jaredallard: Not needed, done later.

shell/ci/auth/gcr.sh
jaredallard: Good concern, planning on doing this later on with all the auth methods. Mostly just supporting a design that can do that in the future right now

PR #96, opened 2021-12-13
fix(hjson-sync): use correct function name

malept:

Fixes the hjson syncer so that it actually runs instead of always failing (but exiting with status 0)


shell/ci/testing/sync-package-hjson.js
```diff
@@ -66,7 +66,7 @@ if (match === null) {
 
         let githubToken;
         try {
-          githubToken = readFileSync(githubTokenPath);
+          githubToken = fs.readFileSync(githubTokenPath);
         } catch (err) {
           console.error("No GitHub token found:", err);
           process.exit();
```
malept, line 72: If no GitHub token is found, perhaps it *should* fail.
```suggestion
          process.exit(1);
```

jaredallard (approved):

PR #103, opened 2021-12-17
feat: DTSS-1095 Allow `test.sh` to run under Delve

richardlarocque:
[DTSS-1095](https://outreach-io.atlassian.net/browse/DTSS-1095)

Updates `test.sh` with an implicit ENV-var parameter called
`PACKAGE_TO_DEBUG`.  When this variable is specified, the `test.sh`
script will invoke Delve on that particular package's tests.

Due to technical limitations, we can only debug one package at a time.
There is no way to run the entire suite of tests under a debugger all at
once, because each pckage creates its own test binary.

### DIFF
--- a/shell/test.sh
+++ b/shell/test.sh
@@ -9,6 +9,11 @@ set -e
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 LINTER="${LINTER:-"$DIR/golangci-lint.sh"}"
 
+# If you want to run tests under a debugger, use this ENV var to specify the
+# name of the package you wish to debug.  You can only debug one package at a
+# time.
+## PACKAGE_TO_DEBUG=
+
 # shellcheck source=./lib/logging.sh
 source "$DIR/lib/logging.sh"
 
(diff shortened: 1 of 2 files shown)


jaredallard:
Can most of this logic be put into `test.sh` and be done via a flag or env var? Seems pretty duplicated.

richardlarocque:
> Can most of this logic be put into `test.sh` and be done via a flag or env var? Seems pretty duplicated.

Yeah, it is.  I was about to write a comment about how "hacks" this PR is, but you beat me to it.  :)

It seems like it should be possible, but there is one big difference between the regular `test.sh` and this one.  We can only debug one single package's tests at a time, where `test.sh` could test all of `./...` at once.  We need an extra argument to specify which package we're debugging.

Would it be reasonable to add a `PKG_TO_TEST` env var?  Should we add actual flags parsing to `test.sh`, so we could have a `--package-to-debug=` argument?  Other ideas?

jaredallard:
Oh, yeah, good point there. I know @jeff-outreach wanted something like this awhile back. I'm totally for modifying that to allow you to specify which packages you'd like to test.

I think argv parsing would be the best way to go, but I'm also hesitant to go back down that path. For now maybe we just go with envvar, and if it's a pain point we can look at better ways to handle this.

richardlarocque:
SGTM.  I'll update it to use an ENV_VAR, merge it with `test.sh`, and ping you when it's fixed.  I'll put this back into draft and remove the review assignment in the meantime.

richardlarocque:
Looks like I can't remove the reviewer assignment.  Oh well.  At least draft-ifying worked.

richardlarocque:
This PR won't win any beauty contests, but it's an improvement on what came before.  I think it's ready for another look.

CC: @jaredallard (when you have time, not urgent)

jeff-outreach (commented):
I have only very general questions and comments about this PR. Where is it documented that this capability will exist and how to use it?  The way devbase is buried in bootstrap makes it non-intuitive to just "discover" these things especially if you don't know they exist.

I'm not against the changes in this PR at all, but just want to note that if you set up the tests cleanly they integrate with the VSCode dlv integration (which is how I have been running tests in the debugger), and it allows running individual tests with breakpoint and inspection integration in VSCode.

Once this goes in (presuming it does) I am curious how it works in comparison to the VS Code integrations.

richardlarocque:
In reply to @jeff-outreach's comment:

In `git` they have the concept of "[plumbing and porcelain](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain)".  This is much more on the "plumbing" side of that taxonomy.

This is not a great public-facing interface.  It's not really intended as such.  Right now, our only "public" interface is the `Makefile` and until that changes it'll be hard to build "porcelain" around something like this.  At best, maybe folks can write integrations with their own preferred editors that build on top of this and contribute them back to bootstrap.

I'm kind of surprised that VSCode's debugger actually works.  It's kind of hard to get a bootstrap project built with all the right build tags and linker flags.  I've struggled to integrate with Emacs' Go tooling, for example.

jaredallard (approved):
LGTM, we should consider documenting this at one point, but it's a great issue right now anyways....

PR #118, opened 2022-01-19
refactor: retry common failure points

jaredallard:

Title says it all.

## Jira ID

[DTSS-1186]

## Notes for your reviewers

[DTSS-1186]: https://outreach-io.atlassian.net/browse/DTSS-1186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/shell.sh
+++ b/shell/lib/shell.sh
@@ -18,7 +18,7 @@ retry() {
 
   local exitCode=0
   for i in $(seq 1 "$maxRetries"); do
-    if [[ $1 -gt 1 ]]; then
+    if [[ $i -gt 1 ]]; then
       echo "RETRYING: $command ($i/$maxRetries)"
     fi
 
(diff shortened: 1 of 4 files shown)


shell/lib/shell.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    # execute command, if succeeds break out of loop and
+    # and reset exitCode
+    "$command" "$@" && exitCode=0 && break
+
+    # preserve the exit code
+    exitCode=$?
+
+    # try again after x time
+    sleep "$interval"
```
george-e-shaw-iv, line 33: could be nice to do sleep interval + (i*3) or something to have a little bit of an exponential backoff

george-e-shaw-iv: "exponential" is not being used literally here

george-e-shaw-iv (approved):

shell/lib/shell.sh
jaredallard: yeah, but i don't think we need it rn, but happy to add it later if this actually becomes an issue (I suspect this'll only ever really trigger once)

PR #119, opened 2022-01-19
fix: Use fixed bundler version (from flagship Gemfile.lock)

plisy:
Different bundler versions create different Gemfile.lock versions, messing things up. This solves that problem.

## Jira ID

[XX-XX]


jaredallard (approved):
Made a note in a private chat but we need to find a way to change this later, given this is OSS

PR #123, opened 2022-01-26
feat: (protoc.sh) update go-protoc-import field to support setting the import path under the module

zaquestion:
this is important particularly when importing from a bootstrap repo. protobuf expects imports to match their generated files

Example:
settings has `table.proto` which it imports as `import table.proto`
settingsworkflow, uses `go-protoc-imports` to import from settings, but the default path required importing table as `import api/table.proto`, these changes allow us to specify that the protoc include path should be under `/api`. This allows the imports to match and the reflection servers to function properly (among avoid other possible mismatch bugs)

Heres what it looks like in a `service.yaml` file:
```
go-protoc-imports:
  - module: "github.com/getoutreach/settings@latest"
    path: "/api"
```

## Notes for your reviewers
- this is a breaking change, however this key is currently only used by settings services all owned by the OLI team. We are happy to bring in as a breaking change and fix manually in our 2 repos

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -32,22 +32,29 @@ if [[ -z $uid ]] || [[ -z $gid ]]; then
 fi
 
 get_import_basename() {
-  basename "$(go list -f '{{ .Path }}' -m "$1" | sed 's|/v[0-9]||') " # sed removes the version off the module path if present
+  local module=$(echo $1 | jq -r .module)
+  # sed removes the version off the module path if present
+  basename $(go list -f '{{ .Path }}' -m "$module" | sed 's|/v[0-9]||')
+}
+get_import_path() {
+  local module=$(echo $1 | jq -r .module)
+  # sed removes the version off the module path if present
+  module_root=$(go list -f '{{ .Dir }}' -m "$module" | sed 's|/v[0-9]||')
+  echo "${module_root}$(echo $1 | jq -r .path)"
 }
 for import in $(get_list "go-protoc-imports"); do
+  module=$(echo $import | jq -r .module)
   currentver="$(go version | awk '{ print $3 }' | sed 's|go||')"
   requiredver="1.16.0"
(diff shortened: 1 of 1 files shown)


jaredallard (approved):

jaredallard (commented):
sidenote: glad I'm not the only person good at bash 😄

shell/protoc.sh
```diff
@@ -32,22 +32,29 @@ if [[ -z $uid ]] || [[ -z $gid ]]; then
...
-  basename "$(go list -f '{{ .Path }}' -m "$1" | sed 's|/v[0-9]||') " # sed removes the version off the module path if present
+  local module=$(echo $1 | jq -r .module)
+  # sed removes the version off the module path if present
+  basename $(go list -f '{{ .Path }}' -m "$module" | sed 's|/v[0-9]||')
+}
+get_import_path() {
+  local module=$(echo $1 | jq -r .module)
+  # sed removes the version off the module path if present
+  module_root=$(go list -f '{{ .Dir }}' -m "$module" | sed 's|/v[0-9]||')
+  echo "${module_root}$(echo $1 | jq -r .path)"
```
jaredallard, line 43:
```suggestion
  echo "${module_root}$(jq -r .path <<< "$1")"
```
useless echo :D

shell/protoc.sh
```diff
@@ -32,22 +32,29 @@ if [[ -z $uid ]] || [[ -z $gid ]]; then
...
 }
 for import in $(get_list "go-protoc-imports"); do
+  module=$(echo $import | jq -r .module)
```
jaredallard, line 46: same here:

```suggestion
  module=$(jq -r .module <<< "$import")
```

jaredallard: (also quoting variables is good practice to avoid accidentally splitting on ` `)

jaredallard:
FYI you can `# shellcheck disable=SC2155` it's not really all that valid IMO and makes bash a lot messier looking. Especially for things like `jq` and what not that you'd be doing empty string checks instead on fail

PR #125, opened 2022-01-26
feat: Don't include or_dev when building in kubernetes

plisy:

When running in devspace, build should not be set up for dev/devenv as that's not available.

## Jira ID

[XX-XX]

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -14,7 +14,7 @@ LDFLAGS             ?= -w -s -X github.com/getoutreach/gobox/pkg/app.Version=$(A
 GOFLAGS             :=
 GOPRIVATE           := github.com/$(ORG)/*
 GOPROXY             := https://proxy.golang.org
-GO_EXTRA_FLAGS      := -v -tags=or_dev
+GO_EXTRA_FLAGS      := -v $(shell [[ ! -z "$KUBERNETES_SERVICE_HOST" ]] && printf "-tags=or_dev")
 TAGS                :=
 BINDIR              := $(CURDIR)/bin
 BIN_NAME            := $(APP)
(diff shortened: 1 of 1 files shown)


root/Makefile
```diff
@@ -14,7 +14,7 @@ LDFLAGS             ?= -w -s -X github.com/getoutreach/gobox/pkg/app.Version=$(A
 GOFLAGS             :=
 GOPRIVATE           := github.com/$(ORG)/*
 GOPROXY             := https://proxy.golang.org
-GO_EXTRA_FLAGS      := -v -tags=or_dev
+GO_EXTRA_FLAGS      := -v $(shell [[ ! -z "$KUBERNETES_SERVICE_HOST" ]] && printf "-tags=or_dev")
```
jaredallard, line 17: nit `! -z` -> `-n`

jaredallard:
```suggestion
GO_EXTRA_FLAGS      := -v $(shell [[ -n "$KUBERNETES_SERVICE_HOST" ]] && printf "-tags=or_dev")
```

jaredallard: Wait, we want `-z` here, right? If `KUBERNETES...` is empty, use `or_dev`. Not the other way around?

jaredallard: or am I misreading the one-liner, I hate bash one-line if statements

plisy: this seems to do what we want, if `KUBERNETES...` is empty, add the `or_dev`. I hate bash.

jaredallard: Hmmmm, ok new suggestion (fun fact don't need to quote in if statement)

ref: https://stackoverflow.com/questions/18096670/what-does-z-mean-in-bash

```suggestion 
GO_EXTRA_FLAGS      := -v $(shell if [[ -z $KUBERNETES_SERVICE_HOST ]]; then printf "-tags=or_dev"; fi)
```

PR #126, opened 2022-01-27
fix: compress yq lines so that protoc-import-changes work

zaquestion:

oops, forgot this was also a required change so that each import can be read as a single line and pulled apart by jq. this change is back compatible with flat lists


george-e-shaw-iv (approved):
i trust u

PR #129, opened 2022-01-31
feat: Extracted build-jsonnet.sh script

plisy:

Extract building jsonnet scripts into separate script so it can be reused in devspace config without the rest of deploy-to-dev.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/build-jsonnet.sh
+++ b/shell/build-jsonnet.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+#
+# wrapper arround jsonnet for rendering files
+SCRIPTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck source=./lib/bootstrap.sh
+source "$SCRIPTS_DIR/lib/bootstrap.sh"
+
+APPNAME="$(get_app_name)"
+
+# shellcheck source=./lib/logging.sh
+source "$SCRIPTS_DIR/lib/logging.sh"
+
+action=$1
+
+bento="${DEVENV_DEPLOY_BENTO:-"bento1a"}"
+namespace="${DEVENV_DEPLOY_NAMESPACE:-$APPNAME--$bento}"
+version="${DEVENV_DEPLOY_VERSION:-"latest"}"
+environment="${DEVENV_DEPLOY_ENVIRONMENT:-"development"}"
+host="${DEVENV_DEPLOY_HOST:-"bento1a.outreach-dev.com"}"
+email="${DEVENV_DEPLOY_DEV_EMAIL:-$(git config user.email)}"
+
+kubecfg \
+  --jurl http://k8s-clusters.outreach.cloud/ \
+  --jurl https://raw.githubusercontent.com/getoutreach/jsonnet-libs/master \
+  -n "$namespace" \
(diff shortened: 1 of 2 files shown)


shell/deploy-to-dev.sh
```diff
@@ -11,10 +11,13 @@ APPNAME="$(get_app_name)"
...
-bento="bento1a"
-version="${DEPLOY_TO_DEV_VERSION:-"latest"}"
-namespace="$APPNAME--$bento"
-environment="${DEPLOY_TO_DEV_ENVIRONMENT:-"development"}"
+
+export DEVENV_DEPLOY_BENTO="bento1a"
+export DEVENV_DEPLOY_VERSION="${DEPLOY_TO_DEV_VERSION:-"latest"}"
+export DEVENV_DEPLOY_NAMESPACE="$APPNAME--$DEVENV_DEPLOY_BENTO"
+export DEVENV_DEPLOY_ENVIRONMENT="${DEPLOY_TO_DEV_ENVIRONMENT:-"development"}"
+export DEVENV_DEPLOY_DEV_EMAIL="${DEV_EMAIL:-$(git config user.email)}"
```
jaredallard, line 19: I think this should be `git config --global user.email`

shell/build-jsonnet.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
jaredallard, line 1: How about instead of making this it's own file we make this `render_jsonnet` function so we can use it in multiple places?

shell/deploy-to-dev.sh
plisy: won't it roll that into local on get?

shell/build-jsonnet.sh
plisy: I need to call it as a script

plisy: it also doesn't help anything much

shell/deploy-to-dev.sh
plisy: yeah, doesn't matter, if current repo has override, it should be used, if it doesn't this will return the global option

jaredallard: Do we want to have the deploy-app functionality change based on the current repo?? I don't think so, also we're going to nuke this method of setting email in the future soon. If we enable local config override that'll make it a breaking change later

shell/build-jsonnet.sh
jaredallard: Alright, though if we have another script that essentially calls this like a fn imma blame you :P

shell/deploy-to-dev.sh
plisy: it's always been this way. I just copied the default. And yeah, you want to do local repo settings if available - think personal vs work repos (shouldn't be very common, but it's a valid scenario)

jaredallard: Well, then we're going to break that later. The idea of dev email is not to be "correct" it's to be a machine wide identifier for metrics/logging tagging, so having it change per repo is not useful.

plisy: There are so many ways that could go wrong ... But for the purpose of this PR this is what we want.

jaredallard (approved):

PR #147, opened 2022-02-23
fix(e2e): set correct TEST_TAGS

jaredallard:

It looks like the `TEST_TAGS` got removed from the Makefile when the E2E rework was done. This was always an issue of misdirection anyways, so now we explicitly set it when we call the E2E runner.

## Jira ID

[XX-XX]


malept (approved):
I tested this with my service, :shipit:

george-e-shaw-iv (approved):

PR #150, opened 2022-03-07
fix(gh-pages): commit changes to the correct folder

malept:

`make docs-publish` wasn't copying docs to the correct folder. GitHub Pages expects it to be in `docs/` (not doing the root folder because that causes far too much headache).

Also only adds that folder to git, to avoid any stray files from ending up in the `gh-pages` branch.


george-e-shaw-iv (approved):
nice 😎

PR #153, opened 2022-03-12
feat: Added support for app image registry override

plisy:

With this change, the image registry passed in by `devenv` will be applied to bootstrap jsonnet files.

See https://github.com/getoutreach/bootstrap/pull/1061

## Jira ID

[DTSS-1456]


jaredallard (approved):
Man, I hate kubecfg.

PR #155, opened 2022-03-22
fix: Use honeycomb apiKey from secret mount if file exists

plisy:

In `devspace` the `apiKey` file is mounted and based on vault secret. There's no need to pull it from vault into a file.

## Jira ID

[XX-XX]


jaredallard (approved):

plisy:
@jaredallard do you see how this can break with different versions of bash/zsh/...? 

I've tested dev container and my regular `orc setup` based machine setup.

PR #156, opened 2022-03-22
fix: Authenticate npm/yarn and bundler in dev container

plisy:
## Jira ID

[XX-XX]

### DIFF
--- a/shell/devspace_start.sh
+++ b/shell/devspace_start.sh
@@ -21,7 +21,13 @@ if [[ -n $GH_TOKEN ]]; then
   echo '//npm.pkg.github.com/:_authToken=${GH_TOKEN}' >>"$HOME/.npmrc"
 
   # We need bundler to be a thing so source ASDF
+  # shellcheck disable=SC1090
+  # shellcheck disable=SC1091
   . "$HOME/.asdf/asdf.sh"
+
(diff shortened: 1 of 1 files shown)


shell/devspace_start.sh
```diff
@@ -7,6 +7,24 @@ source "$DIR/lib/bootstrap.sh"
...
+
+if [[ -n $GH_TOKEN ]]; then
+  # We actually don't want it to expand, we want it to be a literal string written to the file
+  # shellcheck disable=SC2016
+  # shellcheck disable=SC2086
+  echo '//npm.pkg.github.com/:_authToken=${GH_TOKEN}' >>"$HOME/.npmrc"
+
+  # We need bundler to be a thing so source ASDF
+  . "$HOME/.asdf/asdf.sh"
+  bundle config set --global rubygems.pkg.github.com x-access-token:$GH_TOKEN
```
jaredallard, line 25: I _think_ we want `GH_TOKEN` to not evaluate here, right?

plisy: Unfortunately bundle isn't that smart. But the containers are ephemeral enough that it shouldn't matter.

jaredallard (approved):

PR #159, opened 2022-03-24
feat(devspace): Send logs to datadog

plisy:

We pass env variables to help startup script decide whether to start the terminal, or start the app.

These variables are passed from devenv https://github.com/getoutreach/devenv/pull/224

When it starts the app and source watchers, it sends the logs into a file - this file is tailed in container entrypoint. https://github.com/getoutreach/bootstrap/pull/1099

## Jira ID

[DTSS-1520]

## Notes for your reviewers

[DTSS-1520]: https://outreach-io.atlassian.net/browse/DTSS-1520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/devspace_start.sh
+++ b/shell/devspace_start.sh
@@ -66,7 +66,7 @@ This is how you can work with it:
 "
 
 if [[ -z $DEV_CONTAINER_LOGFILE ]] || [[ $DEVENV_DEV_TERMINAL == "true" ]]; then
-  echo "$BANNER"
+  echo -e "$BANNER"
   bash
 else
   make dev | tee -ai "${DEV_CONTAINER_LOGFILE:-/tmp/app.log}"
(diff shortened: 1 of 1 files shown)


shell/devspace_start.sh
```diff
@@ -5,7 +5,7 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 # shellcheck source=./lib/bootstrap.sh
 source "$DIR/lib/bootstrap.sh"
 
-gh auth setup-git
+GH_NO_UPDATE_NOTIFIER=disabled gh auth setup-git
```
jaredallard, line 8: Is this really valid? This is a double negative, I'd think it'd be `GH_UPDATE_NOTIFIER=disabled` or `GH_NO_UPDATE_NOTIFIER=true`

jaredallard: Ah, I see, it's set to any value ([ref](https://cli.github.com/manual/gh_help_environment))

I'd suggest `true` instead so it's easier to read.

```suggestion
GH_NO_UPDATE_NOTIFIER=disabled gh auth setup-git
```

jaredallard (approved):

shell/devspace_start.sh
plisy: I pulled the value out of issues implementing it. Those came up in google first. I can change to true.

PR #161, opened 2022-03-26
feat(devspace): Support debugging non-default binaries (e.g. cdc bins)

plisy:

Enables debugging bins based on environment variable. This is passed through from `devenv`, and bootstrap adds it in `devspace.yaml` config.

Other PRs to make it work e2e:
- https://github.com/getoutreach/bootstrap/pull/1108
- https://github.com/getoutreach/devenv/pull/231

## Jira ID

[DTSS-1466]

## Notes for your reviewers

[DTSS-1466]: https://outreach-io.atlassian.net/browse/DTSS-1466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/debug.sh
+++ b/shell/debug.sh
@@ -13,7 +13,7 @@ export CGO_ENABLED=1
 set -ex
 
 if [[ -n ${DLV_PORT} ]]; then
-  exec "$SCRIPTS_DIR/gobin.sh" github.com/go-delve/delve/cmd/dlv@v"$(get_application_version "delve")" exec "$(get_repo_directory)/bin/$(get_app_name)" --headless --listen=":${DLV_PORT}"
+  exec "$SCRIPTS_DIR/gobin.sh" github.com/go-delve/delve/cmd/dlv@v"$(get_application_version "delve")" exec "$(get_repo_directory)/bin/${DEV_CONTAINER_EXECUTABLE:-$(get_app_name)}" --headless --listen=":${DLV_PORT}"
 else
   exec "$SCRIPTS_DIR/gobin.sh" github.com/go-delve/delve/cmd/dlv@v"$(get_application_version "delve")" debug --build-flags="-tags=or_dev" "$(get_repo_directory)/cmd/$(get_app_name)"
 fi


shell/debug.sh
```diff
@@ -13,7 +13,7 @@ export CGO_ENABLED=1
 set -ex
 
 if [[ -n ${DLV_PORT} ]]; then
-  exec "$SCRIPTS_DIR/gobin.sh" github.com/go-delve/delve/cmd/dlv@v"$(get_application_version "delve")" exec "$(get_repo_directory)/bin/$(get_app_name)" --headless --listen=":${DLV_PORT}"
+  exec "$SCRIPTS_DIR/gobin.sh" github.com/go-delve/delve/cmd/dlv@v"$(get_application_version "delve")" exec "$(get_repo_directory)/bin/${DEV_CONTAINER_EXECUTABLE:-$(get_app_name)}" --headless --listen=":${DLV_PORT}"
```
edmangimelli-outreach, line 16: another dumb question:
that expansion definitely works?
I can't do a contrived version :/
```
 ab= 1 echo "${$(echo a)$(echo b)}"
 # bad substitution
 ```

Comment (plisy) [reply shell/debug.sh:16]: One is in curlies

Comment (plisy) [reply shell/debug.sh:16]: this says, do `$DEV_CONTAINER_EXECUTABLE` if set, otherwise do `$(get_app_name)`

Comment (plisy) [reply shell/debug.sh:16]: in your case the equivalent would be

```sh
a=1 
b=2 
echo "${a:-$(echo $b)}"
```

george-e-shaw-iv (approved):

bhogatemadhura (approved):

shell/debug.sh
edmangimelli-outreach: ah. i missed the hyphen.

boy, bash-style parameter expansion is something i look up probably once a month. can never keep all the variations straight

PR #162, opened 2022-03-30
feat: Enable live reloading

dundee:

Adds support for live reloading of the service using the Air tool. 

Starting the service using `make devserver` every time you do a small change in the code is very time consuming, because it prepares the environment every time (fetches configs, fetches secrets). With live reload, it's a bit faster.

## Jira ID

[XX-XX]


plisy:
Have you seen `make dev`? Is there something that doesn't do and this does?

dundee:
I didn't notice this one 🤣  Thanks!

PR #165, opened 2022-04-04
fix(shell/fmt): skip .snapshot directories

jaredallard:

This PR fixes the fmt script to skip the new snapshot directories but also refactors the invocations of commands to share ignored directories and be less complex to add.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/fmt.sh
+++ b/shell/fmt.sh
@@ -19,26 +19,58 @@ GOFMT="${GOFMT:-gofmt}"
 
 info "Running Formatters"
 
+# for_all_files runs the provided command against all files that
+# match the provided glob. This is powered by find, and thus the glob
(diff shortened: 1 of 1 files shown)


shell/fmt.sh
```diff
@@ -19,30 +19,69 @@ GOFMT="${GOFMT:-gofmt}"
...
+
+    # Skip node modules
+    "*node_modules*"
+  )
+  local glob="$1"
+  shift
+  local command=("$@")
+
+  # create arguments for each directory we're skipping
+  find_args=()
```
jaredallard, line 49:
```suggestion
  local find_args=()
```

george-e-shaw-iv (approved):

PR #168, opened 2022-04-06
feat(orb): move jobs into orbs

jaredallard:

This PR adds a bunch of the jobs currently bootstrap into orb jobs so they can be easily reused in CircleCI.

## Jira ID

[XX-XX]

### DIFF
--- a/orbs/shared/jobs/docker.yaml
+++ b/orbs/shared/jobs/docker.yaml
@@ -1,0 +1,15 @@
+description: Build and (optionally push) a Docker image. Pushes only on release (or default) branches
+executor:
+  name: testbed-machine
+resource_class: xlarge
+steps:
+  - setup_environment:
+      machine: true
+  - build_docker_image
+
+  # IDEA(jaredallard): We'll need to move this out in the future
+  - store_artifacts:
+      path: /tmp/test-results/image_scan.json
+  - run:
+      name: Upload Stats to OpsLevel
+      command: ./scripts/shell-wrapper.sh opslevel/upload.sh
--- a/orbs/shared/jobs/e2e.yaml
+++ b/orbs/shared/jobs/e2e.yaml
@@ -1,0 +1,23 @@
+description: Run E2E tests in a Kubernetes cluster
+parameters:
+  vault_address:
+    description: Vault Instance to use
+    type: string
+    default: ""
+executor:
+  name: testbed-machine
+environment:
+  VAULT_ADDR: << parameters.vault_address >>
+resource_class: large
+steps:
+  - setup_environment:
+      machine: true
+  - with_go_cache
+  - run:
+      name: Run E2E Tests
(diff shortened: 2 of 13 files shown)


orbs/shared/jobs/docker.yaml
```diff
@@ -0,0 +1,15 @@
+description: Build and (optionally push) a Docker image. Pushes only on release (or default) branches
+executor:
+  name: testbed-machine
+resource_class: xlarge
+steps:
+  - setup_environment:
+      machine: true
+  - build_docker_image
+
+  # IDEA(jaredallard): We'll need to move this out in the future
```
george-e-shaw-iv, line 10: `IDEA` followed by the word `need` 🤔

orbs/shared/jobs/e2e.yaml
```diff
@@ -0,0 +1,23 @@
+description: Run E2E tests in a Kubernetes cluster
+parameters:
+  vault_address:
+    description: Vault Instance to use
+    type: string
+    default: ""
```
george-e-shaw-iv, line 6: is the reason we're not providing a default because of oss?

orbs/shared/jobs/docker.yaml
jaredallard: :)

orbs/shared/jobs/e2e.yaml
jaredallard: Yep

jaredallard: This will be passed in by stencil via box config anyways so default will rarely ever be hit

orbs/shared/jobs/test.yaml
```diff
@@ -0,0 +1,102 @@
...
+    description: Localstack version to run
+    type: string
+  localstack_services:
+    default: dynamodb
+    description: Services to run on localstack, comma separated
+    type: string
+
+# DEPRECATED: This will be removed in a future release. e2e should be used
+# for integration+e2e tests, NOT in CircleCI.
+docker:
```
george-e-shaw-iv, line 45: we always have to run all of these? couldn't we just run them conditionally for now

jaredallard: CircleCI doesn't let you conditionally run docker containers :(

george-e-shaw-iv: dumb but ok

george-e-shaw-iv (approved):

orbs/shared/jobs/test.yaml
jaredallard: This will be a breaking change to use in services so this likely won't ever be used in it's current form outside of our projects that will use this

jaredallard: until the next window, that is

PR #169, opened 2022-04-06
fix: only target or_e2e in make e2e

george-e-shaw-iv:
## Jira ID

[XX-XX]

### DIFF
--- a/shell/ci/testing/setup-devenv.sh
+++ b/shell/ci/testing/setup-devenv.sh
@@ -82,5 +82,5 @@ fi
 
 if [[ $E2E == "true" ]]; then
   info "Starting E2E test runner"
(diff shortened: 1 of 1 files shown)


jaredallard:
This _might_ not work from how gobox is setup, also this won't stop us from running normal unit tests because they don't have build tags right now

george-e-shaw-iv:
> This _might_ not work from how gobox is setup, also this won't stop us from running normal unit tests because they don't have build tags right now

What are the gobox concerns? I understand your concerns about test files not having build tags now, but this at least gives people the option to tag their files accordingly to fix the redundancy if they want to.

jaredallard:
> > This _might_ not work from how gobox is setup, also this won't stop us from running normal unit tests because they don't have build tags right now
> 
> What are the gobox concerns? I understand your concerns about test files not having build tags now, but this at least gives people the option to tag their files accordingly to fix the redundancy if they want to.

Chatted via DMs, we don't think gobox will have issues

bhogatemadhura (approved):

PR #172, opened 2022-04-11
feat(orb): replace merge_into_release with merge job for merging a branch into another branch

jaredallard:

This PR changes the `merge_into_release` orb to be `merge` and have generic options instead, so it can be easily reused. Also added a new `push` step that pushes the current branch to the remote, this is now used by `merge`.

## Jira ID

[XX-XX]


deregtd (approved):
chatted offline with feedback.  change or not.  yolo

orbs/shared/jobs/merge.yaml
```diff
@@ -0,0 +1,31 @@
...
+docker:
+  - image: gcr.io/outreach-docker/bootstrap/ci:stable
+    auth:
+      username: _json_key
+      password: $GCLOUD_SERVICE_ACCOUNT
+parameters:
+  head:
+    type: string
+  base:
+    type: string
```
jaredallard, line 13:
```suggestion
  head:
    description: The current branch, this will have base merged into it. This is only supplied to ensure we're on the correct branch when merging.
    type: string
  base:
    description: The branch to merge into head
    type: string
```

PR #179, opened 2022-04-27
feat: Added Telefork API key for CLI telemetry

plisy:
… today

We are going to use Telefork for getting CLI telemetry. For that we need the API key we now store in vault. The API key is public, included in our other clients already.

## Jira ID

[DTSS-1629]

## Notes for your reviewers

[DTSS-1629]: https://outreach-io.atlassian.net/browse/DTSS-1629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -11,7 +11,8 @@ GOOS                ?= $(shell go env GOOS)
 GOARCH              ?= $(shell go env GOARCH)
 PKG                 := $(GO) mod download -x
 HONEYCOMB_API_KEY 	?= $(shell if [[ -f /run/secrets/outreach.io/honeycomb/apiKey ]]; then cat /run/secrets/outreach.io/honeycomb/apiKey; else cat ~/.outreach/$(APP)/honeycomb/apiKey; fi )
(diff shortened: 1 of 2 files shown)


ChrisPRobinson (approved):
LGTM

george-e-shaw-iv:
How are people who can't sign into our vault (anyone who isn't an outreach engineer) going to get the token?

plisy:
> How are people who can't sign into our vault (anyone who isn't an outreach engineer) going to get the token?

How do they get the honeycomb key now?

george-e-shaw-iv:
And we're taking a prerequisite on running devconfig before building CLIs now? Is that already a thing? I assume the idea is to compile them in actually - so my first comment doesn't matter as much but still kind of matters

plisy:
> And we're taking a prerequisite on running devconfig before building CLIs now? Is that already a thing? I assume the idea is to compile them in actually - so my first comment doesn't matter as much but still kind of matters

It's exactly the same thing we do with honeycomb now.

PR #180, opened 2022-04-28
fix: confluence publish with chore commits

artemchernyak-outreach:
This PR ignores any chore commits when looking for changed markdown files. 

## Jira ID

[[DTSS-1567]]

## Notes for your reviewers

Current state: when confluence-publish runs, the actual commit could have one or more chore (bot) commits infront of it. I have found cases where the commit has  a sementic-release-bot in front of it, and a sementic-release-bot and dependabot commit in front. 

This change tries to be a little smarter about which commits we compare by looking at only the non-bot commits.

[DTSS-1567]: https://outreach-io.atlassian.net/browse/DTSS-1567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/ci/release/confluence-publish.sh
+++ b/shell/ci/release/confluence-publish.sh
@@ -75,5 +75,5 @@ while read -r file; do
   else
     info_sub "no space directive found, skipping ${file}"
   fi
-releaseCommits=($(git log --pretty=format:'%H' -n 2 --grep='^chore' --invert-grep))
-done < <(git diff --name-only ${releaseCommits[@]} -- '*.md')
+  readarray -t releaseCommits < <(git log --pretty=format:'%H' -n 2 --grep='^chore' --invert-grep)
+done < <(git diff --name-only "${releaseCommits[@]}" -- '*.md')
--- a/shell/ci/release/confluence-publish.sh
+++ b/shell/ci/release/confluence-publish.sh
@@ -75,5 +75,5 @@ while read -r file; do
   else
     info_sub "no space directive found, skipping ${file}"
   fi
-  previousTag=$(git describe --abbrev=0 --tags "$(git rev-list --tags --skip=1 --max-count=1)" || git rev-list --max-parents=0 HEAD)
+  previousTag=$(git describe --abbrev=0 --tags "$(git rev-list --tags --skip=1 --max-count=1)" || printf '' | git hash-object -t tree --stdin)
 done < <(git diff --name-only HEAD.."$previousTag" -- '*.md')
--- a/shell/ci/release/confluence-publish.sh
+++ b/shell/ci/release/confluence-publish.sh
@@ -75,4 +75,5 @@ while read -r file; do
   else
     info_sub "no space directive found, skipping ${file}"
   fi
-done < <(git diff --name-only HEAD HEAD~1 -- '*.md')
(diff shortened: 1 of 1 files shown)


shell/ci/release/confluence-publish.sh
```diff
@@ -75,4 +75,5 @@ while read -r file; do
   else
     info_sub "no space directive found, skipping ${file}"
   fi
-done < <(git diff --name-only HEAD HEAD~1 -- '*.md')
+  readarray -t releaseCommits < <(git log --pretty=format:'%H' -n 2 --grep='^chore' --invert-grep)
```
jaredallard, line 78: Instead of doing it based on a list of commits, we probably want to change the behaviour to be just on release. Otherwise we'd have to maintain two different lists of what releases/what doesn't.

I'm also concerned about this because we likely have the chance we'd move away from conventional commits in the future in favor of more explicit control of what version should be changed as part of the PR. e.g. specifying "bump: minor" since generally the conventional commit rules have been confusing to people.

jaredallard: Was thinking of a separate issue, this is completely fine with the current grpcClients stuff. So, this is fine.

shell/ci/release/confluence-publish.sh
```diff
@@ -75,4 +75,5 @@ while read -r file; do
...
+done < <(git diff --name-only "${releaseCommits[@]}" -- '*.md')
```
jaredallard, line 79: We _might_ want to consider just doing commits since last tag, but this is fine with me for now

jaredallard (approved):

shell/ci/release/confluence-publish.sh
artemchernyak-outreach: That's a great idea! I can grab the first non-chore commit to represent the current commit and compare it agains the last tag commit.

I also just realized that `git diff` can accept a commit range. Let me get this updated.

artemchernyak-outreach: @jaredallard let me know what you think of the new approach.

shell/ci/release/confluence-publish.sh
```diff
@@ -75,4 +75,5 @@ while read -r file; do
   else
     info_sub "no space directive found, skipping ${file}"
   fi
-done < <(git diff --name-only HEAD HEAD~1 -- '*.md')
+  previousTag=$(git describe --abbrev=0 --tags "$(git rev-list --tags --skip=1 --max-count=1)" || git rev-list --max-parents=0 HEAD)
+done < <(git diff --name-only HEAD.."$previousTag" -- '*.md')
```
george-e-shaw-iv, line 79: i think this soln is vulnerable to the same thing this PR was aiming to prevent - https://github.com/getoutreach/devbase/pull/148

i get where this is going, but we'll need to modify it slightly to also fix what that PR I linked fixes too

artemchernyak-outreach: I learned way more than I ever wanted to know about git hashes, but I think I got a solution that covers all edge cases and isn't too messy.

george-e-shaw-iv (approved):
trust

PR #182, opened 2022-04-29
fix(update-pipeline): use jsonnet from gobin

malept:

`jsonnet` should be run via gobin to ensure that it's installed/used by whatever version of Go that asdf currently has installed.

Also updated the version of `jsonnetfmt` to be consistent. Although at this point I'm not sure where that's used. `shell/fmt.sh` has its own definition:

https://github.com/getoutreach/devbase/blob/8f1d9c437371059f7f94f4904bc5253a6926c6aa/shell/fmt.sh#L15

I'd be happy to delete that line in `root/Makefile` if it's unused.

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -26,7 +26,7 @@ TOOL_DEPS           := ${GO}
 
 # formatters / misc
 CLANG_FORMAT        ?= ./scripts/shell-wrapper.sh clang-format.sh
-JSONNET             ?= ./scripts/shell-wrapper.sh gobin.sh github.com/google/go-/cmd/jsonnet@v0.18.0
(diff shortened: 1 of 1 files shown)


george-e-shaw-iv:
i am in agreeance with you that we should remove `JSONNETFMT` from the makefile - its unnecessary and i tested to confirm that.

however, when i use your branch on oats I get an error when attempting to run `make update-pipeline` (the only thing that uses jsonnet)

did you not get an error?

malept:
I did not when I tried it, but I can try again, in case it was a fluke.

malept:
Deleted the `JSONNETFMT` variable and fixed the invocation of `JSONNET` (should be using `$(CURDIR)/scripts` instead of `./scripts`)

george-e-shaw-iv (approved):

george-e-shaw-iv:
Works :) thanks

PR #188, opened 2022-05-11
fix: update tools version to 1.17.9

ChrisPRobinson:
Fixes the toolversion to 1.17.5 -> 1.17.9 for go in devbase.


ChrisPRobinson:
Not needed this and more in another PR

PR #190, opened 2022-05-11
fix: too many open files for golangci-lint

george-e-shaw-iv:
* this is a fix that it seems has had some success from my tests. the
error will still occur the first time running in large repos such as
dsngsyncworkflow, but subsequent runs w/changed code seem to be fine.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/golangci-lint.sh
+++ b/shell/golangci-lint.sh
@@ -26,7 +26,8 @@ fi
 
 # Use individual directories for golangci-lint cache as opposed to a mono-directory.
 # This helps with the "too many open files" error.
-mkdir -p "$HOME/.outreach/.cache/.golangci-lint" > /dev/null 2>&1
-export GOLANGCI_LINT_CACHE="$HOME/.outreach/.cache/.golangci-lint/$(get_app_name)"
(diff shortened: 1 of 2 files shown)


ChrisPRobinson (approved):
LGTM

shell/golangci-lint.sh
```diff
@@ -24,4 +24,10 @@ if grep "^\d" <<<"$version"; then
   version="v$version"
 fi
 
+# Use individual directories for golangci-lint cache as opposed to a mono-directory.
+# This helps with the "too many open files" error.
+mkdir -p "$HOME/.outreach/.cache/.golangci-lint" >/dev/null 2>&1
+GOLANGCI_LINT_CACHE="$HOME/.outreach/.cache/.golangci-lint/$(get_app_name)"
+export GOLANGCI_LINT_CACHE
```
jaredallard, line 31: nit: can do in the same line

```suggestion
export GOLANGCI_LINT_CACHE="$HOME/.outreach/.cache/.golangci-lint/$(get_app_name)"
```

PR #191, opened 2022-05-12
feat: enable e2e test for library

bhogatemadhura:
## Jira ID

https://outreach-io.atlassian.net/browse/DTSS-1677

## Notes for your reviewers
 `make e2e` locally  runs e2e tests

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -40,6 +40,7 @@ var virtualDeps = map[string][]string{
 // Service is a mock of the service.yaml in bootstrap, which isn't currently
 // open-sourced, yet!
 type Service struct {
+	Library      bool `yaml:"library,omitempty"`
 	Dependencies struct {
 		// Optional is a list of OPTIONAL services e.g. the service can run / gracefully function without it running
 		Optional []string `yaml:"optional"`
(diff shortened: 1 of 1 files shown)


plisy (approved):

e2e/e2e.go
```diff
@@ -152,6 +147,22 @@ func grabDependencies(ctx context.Context, deps map[string]bool, name string, au
 	return nil
 }
 
+func parseServiceYaml(_ context.Context) (*Service, error) {
```
jaredallard, line 150: nit: Don't pass in context if we know we're not going to use it

e2e/e2e.go
```diff
@@ -152,6 +147,22 @@ func grabDependencies(ctx context.Context, deps map[string]bool, name string, au
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read service.yaml")
+	}
+	defer f.Close()
+
+	var s *Service
+	err = yaml.NewDecoder(f).Decode(&s)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse service.yaml")
+	}
```
jaredallard, line 161: nit: Generally don't decode into a pointer, reference it later where needed (which would be the return)

Second part is always try to combine `if err := whatever(); err != nil` to make it easier to handle errors

```suggestion
	var s Service
	if 	err := yaml.NewDecoder(f).Decode(&s); err != nil {
		return nil, errors.Wrapf(err, "failed to parse service.yaml")
	}
```

bhogatemadhura: https://github.com/getoutreach/devbase/pull/193

PR #199, opened 2022-05-20
feat: Add 'go.sh' to run the current project's Go

richardlarocque:
Adds a 'go.sh' script that attempts to infer which is the correct
version of Go for this given project then run it.

Normally `asdf` magically takes care of this by looking at your current
directory and the local `tool-versions` file.  But there are situations
where that doesn't work.  For example, a text editor might be started
from a context where `asdf` doesn't apply, or may have a `$PWD` that
doesn't match the project directory.


richardlarocque:
This is one of those "conversation starter" PRs.  I wanted to offer an example of what I'm trying to achieve as an opening statement.  I'm very much open to suggestions on better ways to do this.

(I feel like `asdf` should have something like this baked in, but unfortunately, there does not seem to be a `BEHAVE_AS_IF_PWD_IS=...` option for it, and I'm guessing just faking out `PWD` could have unexpected consequences.)

vk-outreach:
Hmm.. asdf has the ASDF_DEFAULT_TOOL_VERSIONS_FILENAME env var which lets you point to any specific .tool-verisons file.  I vaguely feel like that may be a more reliable path (as it will work with other binaries too, not just Go).

vk-outreach:
It looks like ASDF_DEFAULT_TOOL_VERSIONS_FILENAME works if the current directory (or its parents) don't have a .tool-versions file.  Not sure if that handles all the use cases you had in mind @richardlarocque

richardlarocque:
> Hmm.. asdf has the ASDF_DEFAULT_TOOL_VERSIONS_FILENAME env var which lets you point to any specific .tool-verisons file. I vaguely feel like that may be a more reliable path (as it will work with other binaries too, not just Go).

Good point.  I'd seen that variable and thought it referred to just the file name, not a full path.

After your comment I looked into it, and it seems it is treated as a full path: https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L221

richardlarocque:
Hm...
```
rlarocque@richard-x1:~/src/mint$ .bootstrap/shell/go.sh version
go version go1.17.9 linux/amd64
rlarocque@richard-x1:~/src/mint$ cd ..
rlarocque@richard-x1:~/src$ ./mint/.bootstrap/shell/go.sh version
go version go1.17.1 linux/amd64
```

That's not what I was hoping for.  I'll debug a bit more.  This current version is not correct.

vk-outreach:
Do you have a .tool-versions in your home dir @richardlarocque ?  I do and that definitely takes precedence over the env var unfortunately.

richardlarocque:
D'oh: https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L219.

Yep, I do have a `.tool-versions` in my `$HOME`.  That would do it, I guess.

I'll revert back to the `ASDF_GOLANG_VERSION`.  It's ugly, but at least it has the right precedence.

```
rlarocque@richard-x1:~/src/mint$ .bootstrap/shell/go.sh version
go version go1.17.9 linux/amd64
rlarocque@richard-x1:~/src/mint$ cd ..
rlarocque@richard-x1:~/src$ ./mint/.bootstrap/shell/go.sh version
go version go1.17.9 linux/amd64
```

richardlarocque:
Ugly version is back up.  It's not ideal, but it should work most of the time, and it's possibly the best approach we can get within the confines of the interface currently provided by `asdf`.

vk-outreach:
This looks fine to me but it makes me uneasy to have something that only works for `go`.  It feels like we should have a `<repo>/bin/exec.sh` that allows executing any command in the context of that repo (i.e. `mint/bin/exec.sh go version` should print the mint go version).  We can implement this just for `go` for now...

But maybe that is overkill and we can stick to the go.sh plan until we have another use case?

richardlarocque:
> This looks fine to me but it makes me uneasy to have something that only works for `go`. It feels like we should have a `<repo>/bin/exec` that allows executing any command in the context of that repo (i.e. `mint/bin/exec go version` should print the mint go version). We can implement this just for `go` for now...
> 
> But maybe that is overkill and we can stick to the go.sh plan until we have another use case?

Fair question.

I'm thinking of this from an API design point of view.  Even if we had a generic "exec" command, would it be weird to have a specialized `go.sh` that points to it?  My opinion is that this would be not too bad.

So I think we're not painting ourselves into a corner here.  We can have a `go.sh` for now, add an "exec" later if and when we need it, and just re-implement `go.sh` on top of that generic exec.

----

I'm reluctant to do the generic exec solution now because I'm not sure exactly how it would work.  This program-specific environment variable seems to be the only way to get `asdf` to do what we want.  I guess we could have an "exec" with a `case` statement on the inside to cycle through a list of supported tools.  I would much prefer a solution that "just works" with any program supported by `asdf` without needing to update that case statement, though.

I'd like to defer the generic exec until we have a solution to that problem.

vk-outreach:
> I'd like to defer the generic exec until we have a solution to that problem.

The solution that I did elsewhere for this was to copy the .tool-versions from the repo temporarily into the current working directory (and restore it afterwards).  Could we do that here?

shell/go.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Parses the embedding project's `.tool-versions` file to find which is the
+# currently active version of Go, then invokes it explicitly through `asdf`.
+#
+# The idea here is that `some_project/.bootstrap/shell/go.sh` will will always
+# point to the "right" go for that particular `some_project`.
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+GOVERSION=$(grep golang "${DIR}/../../.tool-versions" | tail -n 1 | awk '{ print $2 }')
```
jaredallard, line 10: I'd prefer `$(get_repo_directory)` from `lib/bootstrap.sh` instead of `../..`

shell/go.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
jaredallard, line 1: nit: Use `/usr/bin/env bash` over `/bin/bash` to ensure we're using the user's bash version

shell/go.sh
```diff
@@ -0,0 +1,18 @@
...
+# Parses the embedding project's `.tool-versions` file to find which is the
+# currently active version of Go, then invokes it explicitly through `asdf`.
+#
+# The idea here is that `some_project/.bootstrap/shell/go.sh` will will always
+# point to the "right" go for that particular `some_project`.
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+GOVERSION=$(grep golang "${DIR}/../../.tool-versions" | tail -n 1 | awk '{ print $2 }')
+
+if [ -z "${GOVERSION}" ]; then
```
jaredallard, line 12:
```suggestion
if [[ -z "$GOVERSION" ]]; then
```

nit: I prefer to always use double brackets since we're always targeting bash. I'd also suggest _not_ using `${}` unless required. That's a style preference though.

shell/go.sh
jaredallard:
```suggestion
#!/usr/bin/env bash
```

jaredallard:
> But I'm ok with going with go.sh for now... we can try the other options later on.

A reason I'm against _not_ having a `go.sh` is that things that want a binary tend to not be happy with a `exec.sh go.sh` since that's now two arguments. That being said, if we have other cases we can and _should_ abstract this out for the future and make `go.sh` just alias it for sure.

shell/go.sh
```diff
@@ -0,0 +1,18 @@
...
+  echo "Unable to find Go version for this project" >&2
```
george-e-shaw-iv, line 13: isn't it this -
```suggestion
  echo "Unable to find Go version for this project" 1>&2
```
maybe im wrong, not the _best_ with bash

jaredallard: No, `>` implies `1>`

george-e-shaw-iv: nice

richardlarocque: I see what you mean, but I think just `>&2` works.  (I guess the `1` part is implicit?)

https://stackoverflow.com/questions/31656645/how-do-i-echo-directly-on-standard-output-inside-a-shell-function

PR #200, opened 2022-05-20
fix: handle newer commits in confluence publish

artemchernyak-outreach:

This fixes confluence publish to handle an additional case where there is a commit after the current tag. Some publish jobs are failing due to this.

This handles all previous cases plus the case where there was a commit after the current tag.

## Jira ID

[DTSS-1838]

## Notes for your reviewers

[DTSS-1838]: https://outreach-io.atlassian.net/browse/DTSS-1838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/ci/release/confluence-publish.sh
+++ b/shell/ci/release/confluence-publish.sh
@@ -75,5 +75,5 @@ while read -r file; do
   else
     info_sub "no space directive found, skipping ${file}"
   fi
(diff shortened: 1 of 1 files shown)


captain-enjoyable (approved):
Thank you!

PR #203, opened 2022-05-24
feat: no docker for protoc

george-e-shaw-iv:
* get rid of docker use for protoc

## Jira ID

[XX-XX]

## Notes for your reviewers
examples:
* [oats](https://github.com/getoutreach/oats/pull/160/files) (all three clients)
* [settingsworkflow](https://github.com/getoutreach/settingsworkflow/pull/101) (has remote protoc import w/go,ruby clients)


shell/protoc.sh
```diff
@@ -62,25 +54,39 @@ for import in $(get_list "go-protoc-imports"); do
...
+info "Ensuring protoc plugins are installed"
+
+info_sub "protoc-gen-validate"
+go get -d github.com/envoyproxy/protoc-gen-validate@v0.6.7
+go install github.com/envoyproxy/protoc-gen-validate@v0.6.7
 
-# Create a localuser matching our gid and uid to prevent issues with file permissions.
-docker exec "$CONTAINER_ID" sh -c "groupadd -f --gid $gid localuser && useradd --uid $uid --gid $gid localuser"
+info_sub "protoc-gen-go"
+go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
```
jaredallard, line 69: don't use `go install`

jaredallard: you should be able to use gobin in the arguments below, that's preferred

shell/protoc.sh
```diff
@@ -16,74 +11,86 @@ source "$SCRIPTS_DIR/lib/bootstrap.sh"
...
 
 # Create the language specific clients
 info_sub "go"
-# shellcheck disable=SC2046 # Why: We want it to split
-docker exec --user localuser "$CONTAINER_ID" entrypoint.sh -f './*.proto' -l go \
-  $(for import in $(get_list "go-protoc-imports"); do echo "-i /mod/$(get_import_basename "$import")"; done) \
-  $(if has_feature "validation"; then echo "--with-validator --validator-source-relative"; fi) \
-  --go-source-relative -o ./ --with-docs html,index.html
+exec protoc \
+  --plugin=protoc-gen-go=$protoc_gen_go --plugin=protoc-gen-go-grpc=$protoc_gen_go_grpc \
```
jaredallard, line 80: quote yo variables :D

shell/protoc.sh
```diff
@@ -16,86 +11,115 @@ source "$SCRIPTS_DIR/lib/bootstrap.sh"
...
+
+imports="$(get_list "go-protoc-imports")"
 
-# Fallback if uid/gid is somehow empty
-if [[ -z $uid ]] || [[ -z $gid ]]; then
-  echo "Error: Failed to determine the uid/gid of the current user. Defaulting to standard 1000." >&2
-  uid=1000
-  gid=1000
+# Add default imports so they can get dealt with in the following loop just like custom imports.
+if [[ has_feature "validation" ]]; then
```
jaredallard, line 22:
```suggestion
if has_feature "validation"; then
```

jaredallard (approved):

coding-paras (approved):
Overall looks good to me. However, I think this is very bootstrap specific. Is that the goal? For example, the out paths are hardcoded. We (clerk) eventually want to take a dependency on this protoc.sh script and it will be nice if we can pass arguments about:
1. Where to look for things? i.e. where are the proto files
2. Where to put the files?
3. etc.

jaredallard (approved):
i hate this but i also love it. this is conflicting. LGTM

shell/gobin.sh
```diff
@@ -21,7 +21,7 @@ if [[ $1 == "-p" ]]; then
 fi
 
 if [[ -z $1 ]] || [[ $1 =~ ^(--help|-h) ]]; then
-  echo "Usage: $0 [-p|-h|--help] <package> [args]" >&2
+  echo "Usage: [BUILD_DIR=...|BUILD_PATH=...] $0 [-p|-h|--help] <package> [args]" >&2
```
vk-outreach, line 24: I'd be ok with just proxying any args given to gobin.sh over to the underlying gobin....  (instead of us adding these one by one :) )

vk-outreach: Or even do GOBIN_ARGS instead of individual args.

shell/protoc.sh
```diff
@@ -16,86 +11,166 @@ source "$SCRIPTS_DIR/lib/bootstrap.sh"
...
 PROTO_DOCS_DIR="$(get_repo_directory)/apidocs/proto"
 
-if [[ -n $CIRCLECI ]]; then
-  {
-    echo "warning: running protobuf generation in CI is only supported for bootstrap and is DEPRECATED"
-    echo "         this will result in only Go protobuf artifacts being generated"
-  } >&2
-  exec protoc --go_out=plugins=grpc,paths=source_relative:. --proto_path "$(get_repo_directory)/api" "$(get_repo_directory)/api/"*.proto
-fi
+PROTOC_IMPORTS_BASE_PATH="$HOME/.outreach/.protoc-imports"
```
vk-outreach, line 14: Hmm... are there any chances for conflict between protoc versions if they have the same dir?

george-e-shaw-iv: im thinking about it and I don't know if it matters to version this path with the protoc version too. i think it would matter if protoc was compiling the imports separately or something, but it isn't - it's just taking the proto files at face value at whatever revision you specify for that import

vk-outreach: Could we make it work with a random tempdir path?  Which would guarantee a unique path each time.

shell/protoc.sh
```diff
@@ -16,86 +11,166 @@ source "$SCRIPTS_DIR/lib/bootstrap.sh"
...
-    exit 1
+default_args=()
+for import in $imports; do
+  module_version_str=$(jq -r .module <<<"$import")
+  info_sub "$module_version_str"
+
+  # shellcheck disable=SC2206
+  module_version_arr=(${module_version_str//@/ })
+  module="${module_version_arr[0]}"
+  version="latest"
```
vk-outreach, line 34: Is this `latest` check here a change?  I'd really like protoc to be consistent run to run and not depend on any environment at all.  Can we fail if there isn't a version specified?

shell/protoc.sh
```diff
@@ -16,86 +11,166 @@ source "$SCRIPTS_DIR/lib/bootstrap.sh"
...
 
-# Fallback if uid/gid is somehow empty
-if [[ -z $uid ]] || [[ -z $gid ]]; then
-  echo "Error: Failed to determine the uid/gid of the current user. Defaulting to standard 1000." >&2
-  uid=1000
-  gid=1000
+# Add default imports so they can get dealt with in the following loop just like custom imports.
+if has_feature "validation"; then
+  imports="${imports}{\"module\":\"github.com/envoyproxy/protoc-gen-validate@v$(get_application_version "protoc-gen-validate")\",\"path\":\"\"}$'\n'"
 fi
```
vk-outreach, line 24: I'd probably write this as:

```bash
imports = $(
    get_list "go-protoc-imports"
    if has_feature "validation"; then
        cat <<EOF
        {
           "module":  "github.com/envoyproxy/protoc-gen-validate@v$(get_application_version "protoc-gen-validate")",
           "path": ""
        }
EOF
    fi
)     
```

shell/protoc.sh
george-e-shaw-iv: looking at the old code I think it did default to latest because it was using `go install`, but we can definitely error on an unpinned version. since we're releasing this on the 15th of june which is already a breaking change release day this should be fine

PR #211, opened 2022-06-06
fix: releasing

george-e-shaw-iv:
* use outreach version of orb publishing semantic release plugin

## Jira ID

[XX-XX]


jaredallard (approved):

jaredallard:
I hope this doesn't require auth, or that if it does it just works

PR #220, opened 2022-06-11
fix: prefer devenv.yaml over service.yaml

george-e-shaw-iv:
* this was merged into stencil-golang rc earlier

## Jira ID

[XX-XX]

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -148,15 +148,19 @@ func grabDependencies(ctx context.Context, deps map[string]bool, name string, au
 }
 
 func parseServiceYaml() (*Service, error) {
-	f, err := os.Open("service.yaml")
+	f, err := os.Open("devenv.yaml")
+	if err != nil && errors.Is(err, os.ErrNotExist) {
+		f, err = os.Open("service.yaml")
+	}
+
 	if err != nil {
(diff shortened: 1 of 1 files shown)


e2e/e2e.go
```diff
@@ -148,15 +148,19 @@ func grabDependencies(ctx context.Context, deps map[string]bool, name string, au
 }
 
 func parseServiceYaml() (*Service, error) {
```
plisy, line 150: Nit: Should the `Service` type be renamed?

plisy (approved):
YAY, now we can standardize this for flagship too. Sweet.

e2e/e2e.go
george-e-shaw-iv: i left it alone because i figured this should honestly come from a package in `devenv` whenever you finish that ticket you have opened

george-e-shaw-iv: the one for doing this in the devenv natively

george-e-shaw-iv: i would've changed it too if you wanted me to, i just had auto-merge enabled :/

plisy: meh, you're right this will go away when devenv get the implementation

PR #227, opened 2022-06-15
fix: handle deps requiring flagship

jaredallard:
**What this PR does**: Handles the scenario where a dep of a service requires flagship. We shouldn't fail at that level.


plisy (approved):

plisy:
I'd probably prefer running a script over repos that fixes dependencies and submits PRs, but this does the job too.

jaredallard:
> I'd probably prefer running a script over repos that fixes dependencies and submits PRs, but this does the job too.

Eh, we're on a tight schedule as is, it's documented in the breaking changes notes 😄

PR #247, opened 2022-06-29
fix: do not format markdown files

george-e-shaw-iv:
* this causes problems with templated markdown files and results in an infinite back-and-forth [DTSS-2081]

## Jira ID

[XX-XX]

## Notes for your reviewers

[DTSS-2081]: https://outreach-io.atlassian.net/browse/DTSS-2081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/fmt.sh
+++ b/shell/fmt.sh
@@ -78,9 +78,9 @@ info_sub "buf (.proto)"
 info_sub "shfmt (.sh)"
 for_all_files '*.sh' "$SHELLFMTPATH" -w -l
 
-info_sub "prettier (.yaml/.yml/.json/.md)"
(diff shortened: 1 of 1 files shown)


jaredallard:
This is a breaking change technically

george-e-shaw-iv:
> So the better solution to this is figuring out a way to render stencil templates and run all of our formatters (and linters, but thats somewhat unrelated to this problem). If formatting changes any of the templates it should result in a failure

george-e-shaw-iv:
We fixed this using prettierignore

PR #251, opened 2022-07-06
fix: match source relative behavior from original docker-protoc impl

zaquestion:
## Jira ID

[XX-XX]

## Notes for your reviewers

* docker-protoc behavior can be seen here:
https://github.com/namely/docker-protoc/blob/master/all/entrypoint.sh#L149
https://github.com/namely/docker-protoc/blob/master/all/entrypoint.sh#L392
* original implementation of this script using
  `--validator-source-relative`
  https://github.com/getoutreach/devbase/blob/d925a545bf978ebb98174edb25d4828cac5c381c/shell/protoc.sh#L82

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -138,7 +138,7 @@ if has_feature "validation"; then
 
   go_args+=(
     --plugin=protoc-gen-validate="$protoc_gen_validate"
-    --validate_out=lang=go,paths=source_relative:"$(get_repo_directory)/api"
+    "--validate_out=lang=go,paths=source_relative:$(get_repo_directory)/api"
   )
 fi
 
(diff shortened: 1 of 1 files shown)


captain-enjoyable (approved):

zaquestion:
Tested these changes locally against settings to verify the fix, but updating `.bootstrap/shell/protoc.sh` and running `make gogenerate`.

The attempt to add protoc gen validate into the imports was also causing issues like:
```
parse error: Unfinished JSON term at EOF at line 2, column 0
fatal: destination path '/home/zaq/.outreach/.protoc-imports' already exists and is not an empty directory.
jq: error (at <stdin>:1): Cannot index string with string "module"
parse error: Expected string key before ':' at line 1, column 9
```
so I removed that block as well since it's actually unneeded, the `protoc-gen-validate` plugin links in the proto path as part of the plugin itself (wheeee).

PR #257, opened 2022-07-13
feat: skip parameter in shared/test

george-e-shaw-iv:
* optional skip parameter to skip shared/test if supplied

## Jira ID

[XX-XX]


ChrisPRobinson (approved):
LGTM

PR #264, opened 2022-07-25
feat: download box in ci

george-e-shaw-iv:
* adds lib script to download box from remote url in ci

## Jira ID

[XX-XX]

### DIFF
--- a/shell/circleci/setup.sh
+++ b/shell/circleci/setup.sh
@@ -4,6 +4,7 @@ set -e
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 CI_DIR="$DIR/../ci"
+LIB_DIR="$DIR/../lib"
 
 # Ensure that asdf is ready to be used
 echo "🔨 Setting up asdf"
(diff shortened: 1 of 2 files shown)


artemchernyak-outreach (approved):

shell/circleci/setup.sh
```diff
@@ -44,5 +45,7 @@ config:
 storageURL: git@github.com:getoutreach/box
 EOF
 
+"$LIB_DIR/download-box.sh"
```
jaredallard, line 48: Yeah, fair call. I always let tools that need it refresh it since the logic is in go... but that was always going to be a little fragile.

We can remove the above stub now that we're doing this

george-e-shaw-iv: we still need the storageURL which the stub above writes out

george-e-shaw-iv: i think its best if the download script doesn't overwrite what is already there - and instead just looks for a storageURL

jaredallard: Fair, maybe we should just read it from a CircleCI context instead?

george-e-shaw-iv: i think that just further expands the surface area of how we deal with box. i think its best if we kinda keep it in parity with the go stuff as much as possible - and i think this solution is pretty close

PR #266, opened 2022-07-26
fix: be tolerant to file with not compatible build tags

pavelsmejkal:

**What this PR does / why we need it:**
make e2e is failing when there are files with only build tag or_int in the folder with err: "import: no buildable Go source files in ..."
JIRA ID: DT-00

## Jira ID

[XX-XX]


george-e-shaw-iv (approved):
Thanks for this - ran into the same problem just now.

PR #268, opened 2022-07-26
fix: e2e runner skip submodules

george-e-shaw-iv:
## Jira ID

[XX-XX]

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"go/build"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


e2e/e2e.go
```diff
@@ -314,6 +314,13 @@ func shouldRunE2ETests() (bool, error) {
 			return err
 		}
 
+		if info.IsDir() && path != "." {
```
jaredallard, line 317: Would it make more sense to see if it's a valid submodule by checking `.gitmodules` in the root directory?

jaredallard (approved):

e2e/e2e.go
george-e-shaw-iv: maybe

PR #269, opened 2022-07-27
feat(docker-builder): support multiple docker images

jaredallard:

This PR adds the ability to build multiple docker images easily to the docker builder. This is done by reading a `docker.yaml` file in the `deployments/` directory.

The options are well documented inside of the shell script, but essentially this could look like below (for example, engdocs):

```yaml
engdocs:
###Block(customImages)
athens:
  # pushTo is a different image registry to push to, defaults to:
  # (box: .devenv.imageRegistry)/$image (or, if not == appName, $appname/$image)
  pushTo: docker.com/stable/athens
  # extra secrets to expose to the builder, defaults to NPM_TOKEN being exposed
  secrets:
  - id=mySecret,env=MY_SECRET_ENV_VARIABLE
pkgsite:
  # Platforms to build this image for, defaults to linux/amd64,linux/arm64
  platforms:
  - linux/amd64
###EndBlock(customImages)
```

## Jira ID

[DT-0]

## Notes for your reviewers

This was done pretty much just for engdocs, but I know other services also need multiple docker image builds so this should be helpful.

cc @getoutreach/fnd-qss who also may find this useful.


george-e-shaw-iv (commented):
lgtm but im gonna wait until i see the template changes to approve so i can better review potentially

george-e-shaw-iv (approved):

PR #272, opened 2022-07-29
feat: aws region as test env var

george-e-shaw-iv:
* aws region parameter to set AWS_REGION test env var
* built orb and validated it

## Jira ID

[XX-XX]


jaredallard (approved):
Can reveal if we start adding more env vars

PR #275, opened 2022-08-02
fix: add files to git for e2e tests

artemchernyak-outreach:

Adds files to git so that tests can see that test file exits.

## Jira ID

[XX-XX]


jaredallard (approved):
Do you also have to commit it? I'm unsure

artemchernyak-outreach:
The git command that `test.sh` runs seems to pick this up correctly as long as it's added to git.

PR #276, opened 2022-08-02
fix: bump terraform

artemchernyak-outreach:

The older version of terraform does not support Go 1.18. With some service like https://github.com/getoutreach/stork that need to depend on stencil, we have to upgrade them to 1.18 now.

## Jira ID

[XX-XX]


jaredallard:
We can close this because we started using `terraform fmt` from the asdf version instead -- so this no longer relies on the host Go version

PR #282, opened 2022-08-08
feat: do not delete grpc directory for node client

dundee:

We originally wanted to delete this directory so no outdated files would stick there forever...

What appeared to be a good idea in the beginning, turned out to be not so smart after all 😃 
We have to restore several files after every bootstrap or gogenerate run which is not very convenient and not all members of our team got used to it, so we have to comment in every second PR, that these files have to remain there.

## Jira ID

[XX-XX]


jaredallard:
I'm a little confused by this, how is deleting the directory causing problems? Deleting it should just be deleting older protobuf artifacts that would then be re-generated? If we don't delete this directory I think it'd be causing more problems....

dundee:
It's because we don't support importing 3rd party proto files in node/TS clients yet. 

https://outreach-hq.slack.com/archives/C01CAR6GVPD/p1655111046652019

So we have some manually created files there for the client to compile.

jaredallard:
I don't think we want to disable deleting this by default, that's going to make the upgrade story for this a lot harder. Is there any way we can move those into a different directory? Or get the google protobuf proto files imported through the `go-protoc-imports` stanza in the `service.yaml` instead? cc @george-e-shaw-iv

dundee:
Why do you think it's a problem to not delete this directory? It should not influence anybody right away. The worst thing that can happen is that they will have some not needed files there. Golang protobuf stub directory (`/api`) and files are also not deleted on `make gogenerate`. They need to be removed manually if the `.proto` files are removed.

We actually cannot move it because the protoc for node/TS creates relative imports like this:
```
import * as protoapi_protobuf_extensions_pb from "./protoapi/protobuf/extensions_pb";
```
https://github.com/getoutreach/rms/blob/main/api/clients/node/src/grpc/account_pb.d.ts#L14

So it needs to be placed in the `grpc` directory as well.

george-e-shaw-iv:
> Fair, I'm going to let @george-e-shaw-iv have the final say as he designed it to remove this. We chatted about this in person a week ago, but I think it slipped off the radar.

Was likely a fragment of the past when we upgraded a node protoc plugin. You should merge this it'll likely be fine

dundee:
Yes, it was introduced by rms originally, because we thought it's a good idea to have it cleared but we didn't know the consequences yet.

george-e-shaw-iv (approved):

PR #284, opened 2022-08-10
feat: enable rc branch release flow

artemchernyak-outreach:
enabe rc branch release flow


package.json
```diff
@@ -11,8 +11,5 @@
     "prettier": "^2.3.2",
     "semantic-release": "^19.0.0",
     "semver": "^7.3.5"
-  },
-  "dependencies": {
-    "@outreach/semantic-release-circleci-orb": "^1.1.3"
```
jaredallard, line 16: We need this

jaredallard: Ah, yeah, so we do need a block here, but `package.json` doesn't actually support blocks. I want to `fromJSON` this and keep user's changes by basically doing a merge but I haven't gotten around to it yet :(

artemchernyak-outreach: Oh yeah... I forgot about that json annoyance. Oh well, I just added it back as a short term fix. Maybe we can get the `fromJSON` loading working by next release.

service.yaml
```diff
@@ -11,6 +11,7 @@ metrics: datadog
 releaseType: ""
 releaseOptions:
   allowPrereleases: true
+  prereleasesBranch: ""
```
jaredallard, line 14: Should this have a value? 🤔

artemchernyak-outreach: Sorry about that, looks like the PR ran with RC versions. All fixed up now.

jaredallard (approved):

PR #285, opened 2022-08-11
feat: stork config lint rules

Ashvin-Ranjan:
## Jira ID

[DT-2667]

## Notes for your reviewers

[DT-2667]: https://outreach-io.atlassian.net/browse/DT-2667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -226,3 +226,9 @@ benchmarker-gen:
 .PHONY: benchmarker-run
 benchmarker-run:
 	benchmarker --input benchmarker-grpc.json --output benchmarker-output.json --format json --verbose
+
+## check-config:    make sure that stork.yaml and manifest.yaml will not cause erros
+.PHONY: check-config
+check-config:
+	$(BASE_TEST_ENV) CHECK_CONFIG=true
+	./scripts/shell-wrapper.sh ci/testing/setup-devenv.sh
--- a/root/Makefile
+++ b/root/Makefile
@@ -51,7 +51,7 @@ OSS                 ?= false
 default: build
 
 # All the pre-action steps are bunched together here.
-.PHONY: pre-release pre-build pre-lint pre-run pre-test pre-coverage pre-integration pre-e2e pre-benchmark pre-gogenerate pre-devserver pre-dev pre-debug pre-docker-build pre-fmt pre-docs pre-docs-publish pre-check-config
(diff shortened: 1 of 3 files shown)


stork_config/stork_config.go
```diff
@@ -0,0 +1,149 @@
+package main
```
jaredallard, line 1: I'd suggest putting this into the stork CLI, this is an open source repository and it doesn't really belong here.

shell/ci/testing/setup-devenv.sh
```diff
@@ -96,3 +97,8 @@ if [[ $E2E == "true" ]]; then
   info "Starting E2E test runner"
   TEST_TAGS=or_test,or_e2e exec "$("$DIR/../../gobin.sh" -p "github.com/getoutreach/devbase/v2/e2e@$(cat "$DIR/../../../.version")")"
 fi
+
+if [[ $CHECK_CONFIG == "true" ]]; then
```
jaredallard, line 101: Instead of putting this here, you should put this into the `shell/linters` directory and use the linter framework there

root/Makefile
```diff
@@ -226,3 +226,9 @@ benchmarker-gen:
 .PHONY: benchmarker-run
 benchmarker-run:
 	benchmarker --input benchmarker-grpc.json --output benchmarker-output.json --format json --verbose
+
+## check-config:    make sure that stork.yaml and manifest.yaml will not cause erros
```
jaredallard, line 230: This should be apart of the `lint` target

jaredallard (changes requested):

stork_config/stork_config.go
Ashvin-Ranjan: Stork doesn't have a dedicated CLI and I really do not believe this belongs where the current CLI is.

jaredallard: Just throw a linter into the stork repository as a binary, it's much more related to that repository than it is to `devbase` which is really just for common utilities between repositories

Ashvin-Ranjan: Ok, I will put it there then.

PR #287, opened 2022-08-19
fix: misc RMS fixes, localizer cleanup, jq error

jaredallard:

This PR fixes a few misc RMS team issues:

 * Localizer wouldn't always be killed when `/var/run/localizer.sock` was left over from a previous process, this is now detected and cleaned up.
 * `devconfig.sh` would sometimes output a false `jq` error, this was because `deploy-to-dev.sh` had more output added to it. Now we use the dedicated `build-jsonnet.sh` script instead which only returns the YAML.

## Jira ID

[XX-XX]

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -16,7 +16,6 @@ import (
 	"github.com/getoutreach/gobox/pkg/sshhelper"
 	localizerapi "github.com/getoutreach/localizer/api"
 	"github.com/getoutreach/localizer/pkg/localizer"
-	"github.com/getoutreach/orgapi/api"
 	"github.com/go-git/go-billy/v5/memfs"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/storage/memory"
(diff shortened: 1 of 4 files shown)


e2e/e2e.go
```diff
@@ -242,7 +242,38 @@ func provisionNew(ctx context.Context, deps []string, target string) error {
 	return nil
 }
 
+// ensureRunningLocalizerWorks check if a localizer is already running, and if it is
+// ensure it's working properly (responding to pings). If it's not, remove the socket.
+func ensureRunningLocalizerWorks(ctx context.Context) error {
```
george-e-shaw-iv, line 247: Don't we already have these two functions written in a public devenv package? I swore it was one of the first things I did when I joined DT

george-e-shaw-iv (approved):

e2e/e2e.go
jaredallard: It's in localizer itself. I'd go clean up this code to make that shared if we weren't killing it in not too far from now

PR #289, opened 2022-08-19
feat: add check_config to linters

Ashvin-Ranjan:
## Jira ID

[DT-2721]

## Notes for your reviewers

[DT-2721]: https://outreach-io.atlassian.net/browse/DT-2721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


george-e-shaw-iv (approved):

shell/linters/check-config.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+# Linters for stork.yaml and manifest.yaml
+CHECK_CONFIG=$("$DIR/gobin.sh" -p github.com/getoutreach/stork/cmd/check_config@v"$(get_tool_version "check_config")")
```
jaredallard, line 3: nit: because of how this works it'd end up running on all yaml files, we should quickly change the linter format to also support `files=()`, which I can do if you don't have time after today.

Also, by setting `CHECK_CONFIG` here, when this is loaded to determine what extensions/(soon to be files) to run on, we'd end up compiling/downloading this tool. So, we probably want to put this into `check_cfg` or `linter` function instead.

george-e-shaw-iv (approved):

PR #296, opened 2022-09-02
feat(golangci-lint): use asdf

jaredallard:

This PR moves us to use asdf for `golangci-lint`, which is marginally helpful for Go upgrades (e.g. 1.19) as they tend to use newer Go versions then we do, which prevents us from upgrading across the board because gobin builds them with the host Go version.

## Jira ID

[XX-XX]


jaredallard:
Fixes issues in lintroller: https://github.com/getoutreach/lintroller

george-e-shaw-iv (approved):

PR #302, opened 2022-09-13
fix(ci/env/asdf): fix race condition in plugin list

jaredallard:

This PR fixes a race condition with the plugin list caching added to `asdf.sh` in the #301, now we change the ordering of the `source` to have this work properly.

## Jira ID

[DT-2808]

## Notes for your reviewers

[DT-2808]: https://outreach-io.atlassian.net/browse/DT-2808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/ci/env/asdf.sh
+++ b/shell/ci/env/asdf.sh
@@ -70,8 +70,8 @@ fi
 # shellcheck source=../../lib/logging.sh
 source "${LIB_DIR}/asdf.sh"
 
-if [[ "$installedAsdf" == "true" ]]; then
-    # Install preloaded versions, usually used for docker executors
+if [[ $installedAsdf == "true" ]]; then
+  # Install preloaded versions, usually used for docker executors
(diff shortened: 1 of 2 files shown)


shell/ci/env/asdf.sh
```diff
@@ -50,8 +48,30 @@ EOF
...
   # shellcheck disable=SC1090
   source "$BASH_ENV"
+}
+
+installedAsdf=false
+
+# Install asdf if it doesn't exist.
+if [[ ! -e "$HOME/.asdf" ]]; then
+  installedAsdf=true
+  init_asdf
```
george-e-shaw-iv, line 58: it doesn't matter but your order here is weird

george-e-shaw-iv (approved):

PR #303, opened 2022-09-14
fix(asdf): if plugins list fails, empty output

jaredallard:

When we first install `asdf`, we have no plugins, so... it errors when we try to list plugins. Nice.

## Jira ID

[DT-2808]

## Notes for your reviewers

[DT-2808]: https://outreach-io.atlassian.net/browse/DT-2808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/asdf.sh
+++ b/shell/lib/asdf.sh
@@ -3,7 +3,7 @@
 
 # asdf_plugins_list stores a list of all asdf plugins
 # this is done to speed up the plugin install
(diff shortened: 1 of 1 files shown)


ChrisPRobinson (approved):
LGTM

PR #322, opened 2022-09-26
docs(rfc): add declarative configuration rfc

jaredallard:

This PR adds an RFC for moving to declarative configuration/better structuring `devbase` in general. This is also the first test of an RFC like system for large scale changes like this.

The idea of an "RFC" is that it will be merged before the work is done on a project, and will contain references to PRs and timelines/plans for a feature. This allows it to gather feedback during the proposal stage, and during implementation for the design to be reviewed/referenced.

[**RENDERED PREVIEW**](https://github.com/getoutreach/devbase/blob/jaredallard/docs/rfc-1-declarative-configuration/rfcs/1-declarative-configuration.md)

### DIFF
--- a/rfcs/1-declarative-configuration.md
+++ b/rfcs/1-declarative-configuration.md
@@ -20,21 +20,23 @@ This proposal replaces environment variable based configuration being used in Ma
 
 ## Motiviation
 
-As we've been writing `devbase`, and other tooling, we've identified that trying to codify all options for tooling to environment variables, bespoke configuration files, and other obscure means has made it difficult for us to provide options and maintain a common set of toolign.
+As we've been writing `devbase`, and other tooling that uses it, we've identified that the current practice of environment variable based configuration as well as opinionated and unchangeable defaults is not flexible enough for using this tooling both outside of stencil-base (where it is primarily used today) and giving our users the flexibility to easily make minor behaviour changes. This has, ultimately created two pain points:
 
-This, combined with a lack of documentation into all of these options, has laregely made `devbase` unconsumable outside of bootstrap/stencil. Thinking of `devbase` as it's own framework to build software makes it more usable in non-stencil projects (like we do already today w/ submodules)
+ * Documentation is hard to write and not easily discoverable
+ * Finding configuration consumption, and configurable areas is difficult
+ * Using `devbase` outside of `stencil-base` projects is not easy (must use sub-modules, and even then it's hard to "plug and play") 
+
 ### Goals
 
- - Documentation for all options in configuration file(s)
+ - All configuration should be written as go structs (as we intend to only write new targets in), and have godoc compatible documentation on each exported field along with examples as needed.
  - Standard location for all configuration for tooling provided by devbase, et. al
- - Documentation mapping 1:1 with Magefile target, with framework to consume/declare it
  - Documentation on how to write Magefile targets, if the framework docs aren't self-explanatory enough
- - Support for usage outside of stencil-base, et. al
+ - Ensure that all targets are "plug-and-play" compatible (e.g. usable by themselves), they shouldn't require stencil modules to be used, and if they do require certain options to work "out of the box" they should be sane defaults and well documented (as per the above goals)
  - Not breaking, while we could do a lot more if we did a breaking release, breaking _all_ of the existing releasing tooling isn't optimal
 
 ### Non-goals
 
- - Expose new functionality. This would be nice to have, but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
+ - Expose new functionality, this would be nice to have but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
  - Migrating `Makefile` to `Magefile`, while it'd be nice to migrate all of them over, that's also an amount of work to do. While moving to `Magefile` would be good, we shouldn't try to move everything to Go at the same time (TL;DR: Wrap shell were needed).
 
 ## Proposal
@@ -79,7 +81,7 @@ The following targets will be available/configurable day one:
  - `e2e`
  - `docker` (includes: `build`, `publish`)
 
-The rest will still be available for compat, but unconfigurable, e.g. `gobuild` will still build go but will not work for non-go repos.
+The rest will still be available for compatibility, but unconfigurable, e.g. `gobuild` will still build go but will not work for non-go repos.
 
 See [Examples](#examples) for an idea of what options will be available/an idea of some of the steps.
 
--- a/rfcs/1-declarative-configuration.md
+++ b/rfcs/1-declarative-configuration.md
@@ -34,7 +34,7 @@ This, combined with a lack of documentation into all of these options, has lareg
 
 ### Non-goals
 
- - Expose new functionality, this would be nice to have but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
+ - Expose new functionality. This would be nice to have, but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
  - Migrating `Makefile` to `Magefile`, while it'd be nice to migrate all of them over, that's also an amount of work to do. While moving to `Magefile` would be good, we shouldn't try to move everything to Go at the same time. (TL;DR: Wrap shell were needed)
 
 ## Proposal
--- a/rfcs/1-declarative-configuration.md
+++ b/rfcs/1-declarative-configuration.md
@@ -1,5 +1,9 @@
 # Declarative Configuration
 
+| Date       | Status   |
+|------------|----------|
+| 2022-09-26 | Draft    |
+
 ## Table of Contents
 
 <!-- toc -->
@@ -7,10 +11,13 @@
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
-- [Proposal](#proposal)
-  - [Implementation Details](#implementation-details)
-  - [Configuration example](#examples)
-- [Test Plan](#testrelease-plan)
+- [Implementation Details](#implementation-details)
+  - [Repository Configuration](#repository-configuration)
+  - [Magefile Targets Implementation Details](#magefile-targets-implementation-details)
+  - [Magefile Targets Available Day One](#magefile-targets-available-day-one)
+  - [Testing Framework](#testing-framework)
+  - [Configuration Examples](#configuration-examples)
+- [Test Plan](#testing-and-release-plan)
 <!-- /toc -->
 
 
@@ -39,13 +46,9 @@ As we've been writing `devbase`, and other tooling that uses it, we've identifie
  - Expose new functionality, this would be nice to have but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
  - Migrating `Makefile` to `Magefile`, while it'd be nice to migrate all of them over, that's also an amount of work to do. While moving to `Magefile` would be good, we shouldn't try to move everything to Go at the same time (TL;DR: Wrap shell were needed).
 
-## Proposal
+## Implementation Details
 
-The high level proposal here is generally covered in the [Summary](#summary) and [Moviation](#motiviation) sections.
-
-### Implementation Details
-
-#### Repository Configuration
+### Repository Configuration
 
 Configuration would live in the root of repository consuming `devbase`, in `.devbase`.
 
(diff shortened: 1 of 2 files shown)


rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+ - `e2e`
+ - `docker` (includes: `build`, `publish`)
+
+The rest will still be available for compat, but unconfigurable, e.g. `gobuild` will still build go but will not work for non-go repos.
+
+See [Examples](#examples) for an idea of what options will be available/an idea of some of the steps.
+
+#### Testing Framework
+
+Testing would be done via Go Testing, enabling us to easily write unit tests. Integration tests would, unforuantely, still need to be figured out as they rely on being ran in macOS/Linux in a predictable fashion
```
jaredallard, line 88: Maybe we should consider writing this to run them all in containers, or something else that'd be more testable...

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Summary
+
+This proposal replaces environment variable based configuration being used in Make+Magefiles in favor of using declarative yaml based configuration.
+
+## Motiviation
+
+As we've been writing `devbase`, and other tooling, we've identified that trying to codify all options for tooling to environment variables, bespoke configuration files, and other obscure means has made it difficult for us to provide options and maintain a common set of toolign.
+
+This, combined with a lack of documentation into all of these options, has laregely made `devbase` unconsumable outside of bootstrap/stencil. Thinking of `devbase` as it's own framework to build software makes it more usable in non-stencil projects (like we do already today w/ submodules)
```
george-e-shaw-iv, line 25:
```suggestion
As we've been writing `devbase` and other tooling we've identified that trying to codify all options for tooling to environment variables, bespoke configuration files, and other obscure means has made it difficult for us to provide options and maintain a common set of tooling. This, combined with a lack of documentation into all of these options, has largely made `devbase` inconsumable outside of bootstrap/stencil. Thinking of `devbase` as its own framework to build software makes it more usable in non-stencil projects (like we do already today w/ submodules).
```

george-e-shaw-iv: im someone who stumbled upon this open source repository (i do not work for outreach)

> (like we do already today w/ submodules)

wat?

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+### Goals
+
+ - Documentation for all options in configuration file(s)
+ - Standard location for all configuration for tooling provided by devbase, et. al
+ - Documentation mapping 1:1 with Magefile target, with framework to consume/declare it
+ - Documentation on how to write Magefile targets, if the framework docs aren't self-explanatory enough
+ - Support for usage outside of stencil-base, et. al
+ - Not breaking, while we could do a lot more if we did a breaking release, breaking _all_ of the existing releasing tooling isn't optimal.
```
george-e-shaw-iv, line 33:
```suggestion
 - Not breaking, while we could do a lot more if we did a breaking release, breaking _all_ of the existing releasing tooling isn't optimal
```
when u make a list either use end of sentence punctuation or dont but don't mix the two

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+This proposal replaces environment variable based configuration being used in Make+Magefiles in favor of using declarative yaml based configuration.
+
+## Motiviation
+
+As we've been writing `devbase`, and other tooling, we've identified that trying to codify all options for tooling to environment variables, bespoke configuration files, and other obscure means has made it difficult for us to provide options and maintain a common set of toolign.
+
+This, combined with a lack of documentation into all of these options, has laregely made `devbase` unconsumable outside of bootstrap/stencil. Thinking of `devbase` as it's own framework to build software makes it more usable in non-stencil projects (like we do already today w/ submodules)
+### Goals
+
+ - Documentation for all options in configuration file(s)
```
george-e-shaw-iv, line 28: I keep having an issue with how vague the word options is

jaredallard: It's somewhat intentionally vague as it's different dependent on the target, I could all it configuration instead? However that seems equally vague.

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+ - Standard location for all configuration for tooling provided by devbase, et. al
+ - Documentation mapping 1:1 with Magefile target, with framework to consume/declare it
+ - Documentation on how to write Magefile targets, if the framework docs aren't self-explanatory enough
+ - Support for usage outside of stencil-base, et. al
+ - Not breaking, while we could do a lot more if we did a breaking release, breaking _all_ of the existing releasing tooling isn't optimal
+
+### Non-goals
+
+ - Expose new functionality, this would be nice to have but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
+ - Migrating `Makefile` to `Magefile`, while it'd be nice to migrate all of them over, that's also an amount of work to do. While moving to `Magefile` would be good, we shouldn't try to move everything to Go at the same time. (TL;DR: Wrap shell were needed)
```
george-e-shaw-iv, line 38:
```suggestion
 - Migrating `Makefile` to `Magefile`, while it'd be nice to migrate all of them over, that's also an amount of work to do. While moving to `Magefile` would be good, we shouldn't try to move everything to Go at the same time (TL;DR: Wrap shell were needed).
```

punctuation goes after parenthesis at the end of a sentence

rfcs/1-declarative-configuration.md
artemchernyak-outreach: Maybe this should be comments for all options in config files. I think keeping this info as close to the config as possible will have a big benefit.

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+ - Documentation for all options in configuration file(s)
+ - Standard location for all configuration for tooling provided by devbase, et. al
+ - Documentation mapping 1:1 with Magefile target, with framework to consume/declare it
+ - Documentation on how to write Magefile targets, if the framework docs aren't self-explanatory enough
+ - Support for usage outside of stencil-base, et. al
+ - Not breaking, while we could do a lot more if we did a breaking release, breaking _all_ of the existing releasing tooling isn't optimal
+
+### Non-goals
+
+ - Expose new functionality, this would be nice to have but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
```
george-e-shaw-iv, line 37:
```suggestion
 - Expose new functionality. This would be nice to have, but would balloon the amount of work. We're targeting 1:1 compatibility with features/configuration already exposed today, with nice-to-haves being limited.
```

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+As we've been writing `devbase`, and other tooling, we've identified that trying to codify all options for tooling to environment variables, bespoke configuration files, and other obscure means has made it difficult for us to provide options and maintain a common set of toolign.
+
+This, combined with a lack of documentation into all of these options, has laregely made `devbase` unconsumable outside of bootstrap/stencil. Thinking of `devbase` as it's own framework to build software makes it more usable in non-stencil projects (like we do already today w/ submodules)
+### Goals
+
+ - Documentation for all options in configuration file(s)
+ - Standard location for all configuration for tooling provided by devbase, et. al
+ - Documentation mapping 1:1 with Magefile target, with framework to consume/declare it
+ - Documentation on how to write Magefile targets, if the framework docs aren't self-explanatory enough
+ - Support for usage outside of stencil-base, et. al
```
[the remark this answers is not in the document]
artemchernyak-outreach: To add to this, if we plan on making this usable outside of stencil, devbase needs its own init script (which can be stencil based). That script can take care of dropping config files with default configs in the right folder and adding other things.

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+
+The high level proposal here is generally covered in the [Summary](#summary) and [Moviation](#motiviation) sections.
+
+### Implementation Details
+
+#### Repository Configuration
+
+Configuration would live in the root of repository consuming `devbase`, in `.devbase`.
+
+Each configuration file would match the relevant `mage <target>` target, e.g.
```
artemchernyak-outreach, line 50: If we are going to continue to support `make`, for the time being, this should state `make/make <target>`.

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+
+```bash
+$ ls .devbase
+build.yaml
+e2e.yaml
+docker.yaml
+```
+
+**Note**: For shared configuration, e.g. `docker` (which could be shared between `docker-build` and `docker-push`) that's also allowed and can become a unified target.
```
george-e-shaw-iv, line 59: You should come up with a paradigm here. Let's take the same example with `docker-build` and `docker-push` sharing configuration in `docker.yaml`

Maybe the paradigm is this - for these targets with shared config they need to all be postfixed with `docker_` and you write shared config in the root of the yaml document, with config pertaining to that specific target under a key with the same name as what follows the prefix.

So what you'd have is actually `docker_build` and `docker_push` with a `docker.yaml` file that looks like this:
```yaml
i_am: "shared config"
here_is_more: "shared config"
build:
  more: "config"
push:
  moar: "config"
```

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Magefile Targets Implementation Details
+
+Each magefile target that consumes this configuration will live inside of the `devbase` repository in the `targets/<targetName>` package, which will be pulled in by `root/mage.go`
```
george-e-shaw-iv, line 63:
```suggestion
Each magefile target that consumes this configuration will live inside of the `devbase` repository in the `targets/<targetName>` package, which will be pulled in by `root/mage.go`.
```

rfcs/1-declarative-configuration.md
artemchernyak-outreach, line 59: I could see having multiple ways to share configs quickly getting out of had. One way to handle this would be to have a single shared variable file and have a hierarchy where all targets get the shared file, but can supersede these options in their specific config files.

This might result in a big shared variable file, but it keeps things simple until we need the complexity.

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+
+The reasoning for a package per target is to make it easier to test these in isolation, as well as (generated) documentation to be localized to the packages (targets) themselves.
+
+A library will be provided at `pkg/targets`, which will provide functionality to read configuration, and other shared functions between targets.
```
george-e-shaw-iv, line 67:
```suggestion
Each magefile target that consumes this configuration will live inside of the `devbase` repository in the `targets/<targetName>` package, which will be pulled in by `root/mage.go`. The reasoning for a package per target is to make it easier to test these in isolation, as well as (generated) documentation to be localized to the packages (targets) themselves. A library will be provided at `pkg/targets`, which will provide functionality to read configuration and other shared functions between targets.
```

u have three related sentences here, each acting as their own paragraph

rfcs/1-declarative-configuration.md
```diff
@@ -0,0 +1,127 @@
...
+    - id=mySecret,env=MY_SECRET_ENV_VARIABLE
+  # Platforms to build this image for, defaults to linux/amd64,linux/arm64
+  # See: https://github.com/docker/buildx#building-multi-platform-images
+  platforms:
+    - linux/amd64
+ ```
+
+## Test/Release Plan
+
+The test plan for this is to ensure that each target has unit, and where possible, integration tests. Releasing would follow the standard DT team releasing process, which is to be released in the next release window, while being put on the `rc` channel to be consumed by Outreach internally first.
```
george-e-shaw-iv, line 126:
```suggestion
The test plan for this is to ensure that each target has unit and, where possible, integration tests. Releasing would follow the standard DT team releasing process, which is to be released in the next release window, while being put on the `rc` channel to be consumed by Outreach internally first.
```

rfcs/1-declarative-configuration.md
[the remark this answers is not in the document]
jaredallard: I want defaults to be in-code, I don't want configuration to be required unless it's needed.

`docker.yaml` is a good example, the entirety of the configuration works OOTB, you just have to provide a key for the docker image name. If you stray from those defaults, you can override them.

rfcs/1-declarative-configuration.md
george-e-shaw-iv: the reason i did underscore instead of hyphen is because then u can still use hypen if need multiple words after the postfixed part

jaredallard: PTAL at the new wording that expresses what I meant to say here

jaredallard: Talked offline, updated this note to be more clear

rfcs/1-declarative-configuration.md
jaredallard: PTAL, I updated this to be more clear about what I'm envisioning

rfcs/1-declarative-configuration.md
artemchernyak-outreach: We could also nest this arbitrarily. e.g. every config file has a nested structure.


For example, `docker.yaml` would have only the config for all docker targets:
```yaml
docker:
  i-am-shared-for-docker-targets: "config"
```

Then `docker-build.yaml` would have the config for build:
```yaml
docker:
  build:
    i-am-only-used-in-build: "config"
```

That could also allow to easily overriding a config from `docker.yaml` inside `docker-build.yaml`.

rfcs/1-declarative-configuration.md
artemchernyak-outreach: Big fan of the new direction of levering Go docs.

rfcs/1-declarative-configuration.md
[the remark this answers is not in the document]
artemchernyak-outreach: Totally makes sense. If we have the documentation available externally as Go docs, this approach works perfectly.

artemchernyak-outreach:
Also, I think this RFC process works really well.

PR #337, opened 2022-10-04
docs(rfc): daily ci cache rebuilds

jaredallard:

This PR adds a RFC for creating daily ci caches instead of a creating caches on each run.

[**RENDERED PREVIEW**](https://github.com/getoutreach/devbase/tree/jaredallard/docs/rfc-2-ci-caching/rfcs/337-daily-ci-cache-rebuilds.md)

## Jira ID

[DT-2489]

## Notes for your reviewers

[DT-2489]: https://outreach-io.atlassian.net/browse/DT-2489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rfcs/337-daily-ci-cache-rebuilds.md
```diff
@@ -0,0 +1,90 @@
...
+only a problem for a day. This is further mitigated by being able to re-trigger the
+cache generation process anytime.
+
+## Alternatives
+
+- Keep the current cache generation process, and just fix the cache generation to
+  properly include all directories required for caching (e.g. go cache). This would
+  help make the cache more efficient, but it would not solve the problem of the asdf
+  version cache being bloated over time without introduction of more dynamic cache
+  key computation and would continue to bloat CI logs.
```
artemchernyak-outreach, line 90: I really like keeping the current process. I think generating the cache daily could be problematic, as you outlined.

What about keeping both options. Upload a new cache if things change in a job but also use a daily cache run for pruning. The daily flow would stay the same as you described, but the goal would change to removing things that we won't need for future runs. We could even reduce the frequency of this since it's only goal is to reduce bloat.

jaredallard: @artemchernyak-outreach I don't think I like that. Mainly because cache generation bloat is still in the CI steps display, and now we have the added complexity of having to run the steps to regenerate the cache in two different places (pruning _is_ regenerating it, ultimately). I think being able to trigger it whenever you want solves pretty much all of the issues around invalidating it, which is what doing it on each job kinda does for you.

george-e-shaw-iv (commented):
this feels incomplete without answering the following questions clearly:
* how many different cached envs do we need to generate per day
* what all are we putting into each cache
  * conversely, what are we purposefully leaving out <- this can be a bit generic but there may be some big gotchas here to expose/callout

rfcs/337-daily-ci-cache-rebuilds.md
artemchernyak-outreach: Talked about this during standup. And came to the conclusion:

If we only run the daily job conditionally (based on new changes to the repo) we will proceed with the implementation of the daily job to update the cache.

rfcs/337-daily-ci-cache-rebuilds.md
```diff
@@ -0,0 +1,114 @@
...
+
+- **Machine Mode Only** `~/.asdf` - The asdf version manager language installs and go cache
+- `~/go/pkg` - The Go package cache (broken, this is per asdf version (see `go env GOROOT`))
+- `~/.outreach/.cache/.golangci-lint` - golangci-lint cache
+- `~/.cache/go-build` - The Go build cache (broken, we don't output build cache there (see `go env GOCACHE`))
+
+The planned cache will instead be:
+
+- **Now Always** `~/.asdf` - The asdf version manager language installs, etc
+- `~/.outreach/.cache` - All caches for Outreach tooling (including `golangci-lint`)
```
artemchernyak-outreach, line 80: Why would we put `.golangci-lint` cache with outreach specific caches? Seems like it fits in better with the generic caches.

artemchernyak-outreach: Up to you if you want to make the move part of this RFC or not. I can see either one being valid. But I do think that we should move it sooner or later.

artemchernyak-outreach (approved):

PR #340, opened 2022-10-06
feat: protoc multiple directories support

4apkis:

Mainly, if we need to generate go classes, the first thing we have to do - provide `go:generate` instructions, like this
`//go:generate ../scripts/shell-wrapper.sh protoc.sh` 
The problem is that at this point this script will search for proto files in a hardcoded api directory only. It's pretty weird, and thus each time developer put proto file to other package, we need a workaround and cannot rely on make gogenerate

As long as protoc compiler does not support recursive search in a nested directories, we need to pass `-proto_path` as well as `--go-out` for the input and output. Mainly we can use repository directory `$(get_repo_directory)` for both, input and output and provide an absolute path to the actual proto file using `pwd` function. This way protoc will detect the package and generate go files in exact same directory where proto files are located (of course, if developer did not forget to provide `go:generate` instructions in the same package)

## Jira ID

[XX-XX]

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -115,14 +115,15 @@ go_args+=(
   --plugin=protoc-gen-go="$protoc_gen_go"
   # --plugin=protoc-gen-go-grpc="$protoc_gen_go_grpc"
   # --go_out=.
-  --go_out=plugins=grpc:. # Remove this line when we upgrade golang protobuf generation (and uncomment everything else).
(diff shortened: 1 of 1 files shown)


jaredallard:
As it stands this would be a breaking change, so we don't plan on taking this on yet. Please let's chat about this in #dev-tooling-design first. Thank you!

PR #341, opened 2022-10-06
fix: stencil module

jaredallard:

This PR turns this repository into a stencil module so that it can be versioned like one.

## Jira ID

[DT-0]


templates/scripts/devbase.sh.tpl
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# This script ensures we have a copy of devbase
+set -e
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+libDir="$DIR/../.bootstrap"
```
artemchernyak-outreach, line 6: Should we switch this to be `.devbase`?

jaredallard: Yes, but we can't right now because I know for a fact that things have hardcoded that path 😢

artemchernyak-outreach (approved):

PR #346, opened 2022-10-10
fix(tool-versions): ruby 2.7.5

jaredallard:

Backports ruby 2.7.5 to devbase `2.2.x`, which is the [version being used by bootstrap v10.2.x](https://github.com/getoutreach/bootstrap/blob/v10.2.2/versions.yaml#L21). 

For posterity, I created `release-2.2` by doing the following:

```bash
$ git checkout v2.2.1
$ git checkout -b release-2.2
```

```yaml
# Updated .releaserc.yml to be the following:
branches:
- name: release-2.2
  range: 2.2.x
  channel: ''
```

**Note**: `yarn run semantic-release` doesn't work until you push the branch up to the remote, sadly.

## Jira ID

[XX-XX]

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -2,7 +2,7 @@
 golang 1.17.9
 nodejs 16.13.0
 terraform 0.13.5
-ruby 2.6.6
+ruby 2.7.5
 protoc 3.19.1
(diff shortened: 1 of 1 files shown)


jkinkead (approved):
Change LGTM!

PR #362, opened 2022-10-26
feat(make): add a test-e2e command

artemchernyak-outreach:

Adds a command for running e2e tests inside a session of `devenv apps e2e`. This runs e2e tests without a linter which should help with general performance of e2e tests in devspace.

## Jira ID

[DT-2967]

## Notes for your reviewers

[DT-2967]: https://outreach-io.atlassian.net/browse/DT-2967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -116,16 +116,16 @@ lint:: pre-test pre-lint
 test:: pre-test lint
 	$(BASE_TEST_ENV) ./scripts/shell-wrapper.sh test.sh
 
+## test-e2e:        run only e2e test (use inside a dev pod)
+.PHONY: test-e2e
+test-e2e:: pre-test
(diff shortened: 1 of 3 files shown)


root/Makefile
```diff
@@ -122,6 +122,11 @@ coverage:: pre-coverage
 	 WITH_COVERAGE=true GOPROXY=$(GOPROXY) GOPRIVATE=$(GOPRIVATE) ./scripts/shell-wrapper.sh test.sh
 	 go tool cover --html=/tmp/coverage.out
 
+## test-e2e:        run only e2e test (use inside a dev pod)
```
jkinkead, line 125: Nit: I think this would be easier to understand if it was under `test` instead of right above `e2e`.

jkinkead (approved):

PR #365, opened 2022-10-31
fix(docker): properly read image specific options

jaredallard:

`name` was never set, now it is

## Jira ID

[DT-3057]


tonymbugua-outreach (approved):
LGTM - adds specific options

PR #369, opened 2022-11-10
fix(gobuild): ensure we have secrets for build

artemchernyak-outreach:

Since we still embed honeycomb and telefork keys in the binaries, this ensure that we have the keys available when we run `make build`.

## Jira ID

[DT-2717]

## Notes for your reviewers

[DT-2717]: https://outreach-io.atlassian.net/browse/DT-2717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -144,6 +144,8 @@ gogenerate:: pre-gogenerate
 ## gobuild:         build application binary
 .PHONY: gobuild
 gobuild:
+	# ensure that we have secrets for the build phase
+	if [[ -z $$SKIP_DEVCONFIG ]]; then ./scripts/shell-wrapper.sh devconfig.sh; fi
 	@$(MAGE_CMD) gobuild
 
 ## grpcui:          run grpcui for an already locally running service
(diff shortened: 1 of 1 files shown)


root/Makefile
```diff
@@ -144,6 +144,8 @@ gogenerate:: pre-gogenerate
 ## gobuild:         build application binary
 .PHONY: gobuild
 gobuild:
+	# ensure that we have secrets for the build phase
+	if [[ -z $$SKIP_DEVCONFIG ]]; then ./scripts/shell-wrapper.sh devconfig.sh; fi
```
jaredallard, line 148: Ehhhh, the reason we made this a separate step (`devserver`) was because `devconfig.sh` is slow to run. I don't like making user's endure that on every local build

artemchernyak-outreach: It doesn't seem that slow. I don't know if there is any other way to ensure this 100%.

What do you think about making this a little easier by providing `make devconfig`? At least that way it's a little more intuitive what to run.

jaredallard: `make devconfig` is good for now, hopefully this'll all be moot after devspace anyways 🙏

jaredallard (approved):

root/Makefile
artemchernyak-outreach: Unfortunately, devspace still relies on the same mechanism.

PR #377, opened 2022-11-17
fix: restencil to fix e2e

chatchai-outreach:
## Jira ID

[XX-XX]


package.json
```diff
@@ -1,16 +1,15 @@
 {
   "license": "UNLICENSED",
   "devDependencies": {
-    "@outreach/semantic-release-circleci-orb": "^1.1.9",
```
jaredallard, line 4: Need to add this orb back

jaredallard: Make sure to run `yarn` after as well

jaredallard (approved):
LGTM after comment

package.json
chatchai-outreach: should we update stencil-base for a custom block?

jaredallard: Yes, but `package.json` doesn't support blocks. I was going to write merging logic but still haven't got around to it :/

PR #378, opened 2022-11-17
fix: Bumping up markdowntools

masterkidan:

Bumps markdowntools version to add support for blockquotes.

## Jira ID

[XX-XX]


chatchai-outreach (approved):
LGTM

PR #386, opened 2022-11-29
fix: re-stencil 11-29-2022

artemchernyak-outreach:

Switches to stable version of all module and re-stencils. Test of full release.

## Jira ID

[XX-XX]

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
   "license": "UNLICENSED",
   "devDependencies": {
-    "@outreach/semantic-release-circleci-orb": "^1.1.9",
     "@semantic-release/commit-analyzer": "^9.0.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,7 +1,6 @@
 {
   "license": "UNLICENSED",
   "devDependencies": {
-    "@outreach/semantic-release-circleci-orb": "^1.1.9",
```
chatchai-outreach, line 4: looks like we added it [manually](https://github.com/getoutreach/devbase/commit/86f9a4e0dd714d427f6144b83caac5c373926056). should it be retained?

artemchernyak-outreach: Yeah, we do need to keep this. Thanks for the catch!

I am going to close this PR, however, and we can re-stencil with the next version.

PR #388, opened 2022-11-30
feat(prtoc): copy imported pb files

artemchernyak-outreach:

This was the only path to make this change non breaking. This approach simply make the process of copying `.pb` files automatic.

This approach places the files in the same location as repos currently expect them to be. I tried many different approaches to getting things to import from other locations and none seemed to work. The common consensus on the node grpc plugin issues list is to follow the same process as well. 

## Jira ID

[DT-2822]

## Notes for your reviewers

[DT-2822]: https://outreach-io.atlassian.net/browse/DT-2822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -64,6 +64,7 @@ imports=$(
 )
 
 default_args=()
+import_paths=()
 for import in $imports; do
   module_version_str=$(jq -r .module <<<"$import")
   info_sub "$module_version_str"
@@ -91,6 +92,7 @@ for import in $imports; do
 
   # Add import to list
   default_args+=("--proto_path=$import_path/$path")
+  import_paths+=("$import_path/$path")
 
   # Check to see if we already have this version locally and skip cloning
   # if we already do.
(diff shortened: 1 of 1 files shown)


george-e-shaw-iv (commented):
are we going to .gitignore these or have service owners commit them?

george-e-shaw-iv:
this also begs the question - should we just be doing this for ruby and go too? we're duplicating data now and resolving .proto dependencies differently between ruby/go and node now

artemchernyak-outreach:
@george-e-shaw-iv Yes, I was planning to open a separate PR to update the `.gitignore` file to ignore that folder.

As far as this diverging from the ruby an Go process, that's just us dealing with the realities of the generation tooling. We have not receive any reports of issues with Go or Ruby, and based on some testing, both seem to handle imports correctly without needing the files to exist in the same folder.

This seems to be an isolated issue to the node grpc plugin and therefore it require special handling. You can see more details here https://github.com/agreatfool/grpc_tools_node_protoc_ts/issues/67.

artemchernyak-outreach:
One correction (shouldn't answer so late at night). We do expect people to check these in. That's the current process and this should not change it in any way. As a test for this, I tested it with https://github.com/getoutreach/commitbridge/tree/main/api/clients/node/src/grpc. They import `orgapi` and if you run `make gogenerate` without these changes, it remove all the `orgapi` files from that directory. With these changes, there is no diff.

pbr0ck3r (approved):

jaredallard (approved):

PR #393, opened 2022-12-16
fix: clone protoc import target everytime

zaquestion:
ensures that when tagging to a branch (such as for developing) that it is always up to date during generation

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -97,7 +97,7 @@ for import in $imports; do
   # Check to see if we already have this version locally and skip cloning
   # if we already do.
   if [[ -d $import_path && -d "$import_path/.git" ]]; then
-    continue
+      rm -r "$import_path"
   fi
 
   # Clone import into the import path since we don't have it already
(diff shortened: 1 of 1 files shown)


jaredallard (approved):

shell/protoc.sh
```diff
@@ -97,7 +97,7 @@ for import in $imports; do
   # Check to see if we already have this version locally and skip cloning
```
jkinkead, line 97: Fix docs?

shell/protoc.sh
```diff
@@ -101,10 +101,10 @@ for import in $imports; do
   default_args+=("--proto_path=$import_path/$path")
   import_paths+=("$import_path/$path")
 
-  # Check to see if we already have this version locally and skip cloning
-  # if we already do.
+  # Check to see if we already have this version locally and remove
+  # to allow a fresh clone if we already do
```
jkinkead, line 105: You're just going to remove the terminal period? Dirty.

zaquestion:
```suggestion
  # to allow a fresh clone if we already do.
```

PR #395, opened 2022-12-23
fix: remove unsupported validate proto plugin

artemchernyak-outreach:

Remove the imports of an unsupported plugin from all generated proto files. We still need to keep the plugin because it's heavily used in Go.

We have proved out that removing this manually still keeps the clients working as expected.

## Jira ID

[DT-3283]

## Notes for your reviewers

[DT-3283]: https://outreach-io.atlassian.net/browse/DT-3283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/protoc.sh
+++ b/shell/protoc.sh
@@ -152,6 +152,10 @@ if has_feature "validation"; then
   )
 fi
 
+delete_validate() {
+  sed -i '' '/validate_pb/d' "$1" 2>/dev/null || sed -i '/validate_pb/d' "$1"
+}
+
 # Make docs output directory if it doesn't exist.
 mkdir -p "$(get_repo_directory)$workDir/doc"
 
@@ -214,10 +218,8 @@ if has_grpc_client "node"; then
 
   # remove unsupported validate proto plugin
   if pushd "$grpc_path"; then
-    for f in * **/*; do
-      if [ -f "$f" ]; then
-        sed -i '' '/validate_pb/d' "$f" 2>/dev/null || sed -i '/validate_pb/d' "$f"
-      fi
+    find . -name '*.js' -or -name '*.ts' | while read -r file; do
+      delete_validate "$file"
     done
     popd || exit 1
   fi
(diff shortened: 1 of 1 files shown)


shell/protoc.sh
```diff
@@ -211,6 +211,16 @@ if has_grpc_client "node"; then
   )
 
   "$grpc_tools_node_bin" "${ts_args[@]}" "$(get_repo_directory)$workDir$SUBDIR/"*.proto
+
+  # remove unsupported validate proto plugin
+  if pushd "$grpc_path"; then
+    for f in * **/*; do
```
jkinkead, line 217: Isn't this the same as `* */*` in bash?

Any reason not to use a `find` command, like `for f in $(find . -name '*.js')` or similar?

artemchernyak-outreach: Not quite. `* **/*` will recursively go as deep as it needs to in the directory tree. `* */*` will only go one level deep.

Generally it's considered bad taste to use `find` for things that glob can handle. Plus everything in this directory should only be auto generated stuff.

jkinkead: This isn't true with the version on macos (5.2):

```bash
bash-5.2$ mkdir -p this/is/a/long/path
bash-5.2$ touch this/is/a/long/path/example.txt
bash-5.2$ cd this/
bash-5.2$ ls **/*
long
bash-5.2$ ls * **/*
is:
a

is/a:
long
bash-5.2$ 
```

artemchernyak-outreach: Oh yeah, I totally forgot about `globstar` being off by default on Macs and older bash versions. This new pattern should handle all the cases.

shell/protoc.sh
```diff
@@ -152,6 +152,10 @@ if has_feature "validation"; then
   )
 fi
 
+delete_validate() {
+  sed -i '' '/validate_pb/d' "$1" 2>/dev/null || sed -i '/validate_pb/d' "$1"
```
jkinkead, line 156: Missing argument to `-i` in second half:
```suggestion
  sed -i '' '/validate_pb/d' "$1" 2>/dev/null || sed -i '' '/validate_pb/d' "$1"
```

jkinkead (approved):
One bug, LGTM otherwise!

shell/protoc.sh
artemchernyak-outreach: The linux version actually doesn't take a second half. That's why this is trying to do this messy `||` logic so it supports both mac and linux.

PR #407, opened 2023-01-14
feat: post e2e script

george-e-shaw-iv:
* runs a script after the devenv has deployed in an e2e test run,
  if it exists.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/ci/testing/setup-devenv.sh
+++ b/shell/ci/testing/setup-devenv.sh
@@ -8,7 +8,7 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 source "$DIR/../../lib/logging.sh"
 
 # shellcheck source=../../lib/bootstrap.sh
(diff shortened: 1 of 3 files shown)


shell/ci/testing/setup-devenv.sh
```diff
@@ -92,6 +96,11 @@ if [[ $PROVISION == "true" ]]; then
   exec devenv --skip-update provision $PROVISION_ARGS
 fi
 
+# If scripts/devenv/post-e2e-deploy.sh exists within the project then run it.
+if [[ -f "$PROJECT_DIR/scripts/devenv/post-e2e-deploy.sh" ]]; then
```
jaredallard, line 100: Can you add documentation to the e2e docs in `docs/makefile.md`? We can get more proper in the docs in the future, but would be good to note that this exists.

Longer term we should move away from shell scripts entirely, but this makes sense with the current state of this repository

jaredallard (approved):

PR #424, opened 2023-02-02
fix: publish all files every release

kaldorn:
There are times when a release fails and docs don't get published, and when a future release happens the older docs (since we compare current diff vs. previous diff) don't get updated. The preferred behavior is to publish all docs to confluence each release, and only update engdocs/godocs on semantic/tagged releases.

## Jira ID

[DT-3205]

## Notes for your reviewers

[DT-3205]: https://outreach-io.atlassian.net/browse/DT-3205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/ci/release/docs.sh
+++ b/shell/ci/release/docs.sh
@@ -9,7 +9,7 @@ LIB_DIR="${DIR}/../../lib"
 source "${LIB_DIR}/logging.sh"
 
 TAG="$CIRCLE_TAG"
-# Do not update engdocs every run
+# Do not update engdocs unless there is a tag
 if [[ -n $TAG ]]; then
   # We need to use the module path to support major versions properly
   MODULE_PATH="$(go list -f '{{ .Path }}' -m)"
--- a/shell/ci/release/docs.sh
+++ b/shell/ci/release/docs.sh
@@ -10,7 +10,7 @@ source "${LIB_DIR}/logging.sh"
 
 TAG="$CIRCLE_TAG"
 # Do not update engdocs every run
-if [[ ! -z "$TAG" ]]; then
+if [[ -n $TAG ]]; then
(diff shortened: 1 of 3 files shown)


shell/ci/release/docs.sh
```diff
@@ -9,20 +9,18 @@ LIB_DIR="${DIR}/../../lib"
...
 TAG="$CIRCLE_TAG"
-if [[ -z $TAG ]]; then
-  # Calculate the psuedo-semver tag, this is used for non-v2 services
-  # (things without semantic-release, generally)
-  TAG="v0.0.0-$(TZ=UTC git --no-pager show --quiet --abbrev=12 --date='format-local:%Y%m%d%H%M%S' --format='%cd-%h')"
-fi
-
-# We need to use the module path to support major versions properly
-MODULE_PATH="$(go list -f '{{ .Path }}' -m)"
+# Do not update engdocs every run
```
jaredallard, line 12: nit: "Don't update engdocs unless there is a tag"

shell/ci/release/docs.sh
```diff
@@ -9,20 +9,18 @@ LIB_DIR="${DIR}/../../lib"
...
 
-info "updating engdocs"
-curl -X POST "$URL"
+  info "updating engdocs"
+  curl -X POST "$URL"
+fi
 
+# Confluence docs get updated every run
 info "publishing eligible markdown documents to confluence"
 "$DIR/confluence-publish.sh"
```
jaredallard, line 26:
```suggestion
exec "$DIR/confluence-publish.sh"
```

jaredallard (approved):

PR #437, opened 2023-02-08
feat: check opslevel status for non service repos

artemchernyak-outreach:

Adds a new mage command to check github actions. 

This is very specific to our repos at this time, but it is written in a way that should be easy to make configurable in the future.

## Jira ID

[DT-2954]

## Notes for your reviewers

[DT-2954]: https://outreach-io.atlassian.net/browse/DT-2954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


root/Makefile
```diff
@@ -238,6 +238,11 @@ benchmarker-run:
 deploy:
 	@$(MAGE_CMD) deploy $(APP) $(CHANNEL)
 
+## checkopslevel: checks if the app passes require opsevel checks
+.PHONY: checkopslevel
```
jaredallard, line 242: The catch-all should do this, right? See L241

jaredallard: I don't know why the deploy one was explicitly added here either tbh...

artemchernyak-outreach: But will it pass through the `$(APP)`? I thought that's why we created specific steps.

artemchernyak-outreach: But will it pass through the `$(APP)`? I thought that's why we created specific steps.

shell/ci/release/release.sh
```diff
@@ -27,6 +27,14 @@ send_failure_notification() {
 # shellcheck source=../../lib/logging.sh
 source "${LIB_DIR}/logging.sh"
 
+# shellcheck source=../../lib/bootstrap.sh
+source "${LIB_DIR}/../../lib/bootstrap.sh"
+
+if [[ "$(is_service)" == "false" ]]; then
```
jaredallard, line 33: This is going to make it so that all CLIs require opslevel. I'd imagine we want to have a config setting in the box config to allow enabling and disabling this, right? Otherwise OSS would require OpsLevel now

shell/lib/bootstrap.sh
```diff
@@ -138,3 +138,11 @@ get_keys() {
     yq -r ".arguments.\"$name\" | keys[]" <"$(get_service_yaml)"
   fi
 }
+
+is_service() {
```
jaredallard, line 142: nit: Add function comment

root/Makefile
jaredallard: Agh. Good point. I hate it :/

jaredallard: Really the target should determine `APP` itself. Now we depend on `make` functionality here.

shell/ci/release/release.sh
artemchernyak-outreach: Yeah... This creates an annoying situation once again. We would ideally turn this off by default and have all internal repos enable this. In the current state, however, there is no easy way for us to enable it for all internal repos.

I am open to ideas here. I wouldn't want to make OSS opt out either though.

root/checkopslevel.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(c.LifecycleToLevelMap) < service.Lifecycle.Index {
+		return false, fmt.Errorf("unsupported lifecycle %d %s",
+			service.Lifecycle.Index, service.Lifecycle.Name)
+	}
+
+	expectedLevelIndex := c.LifecycleToLevelMap[service.Lifecycle.Index]
+	return currentLevelIndex >= expectedLevelIndex, nil
+}
+
+// shouldBeSkipped checks if a service has gating disabled. If it does, it should be skipped.
```
jaredallard, line 134: I left a previous comment here but it somehow nuked itself. We should add unit tests here. Also it seems like `gating_disabled` is a specific tag for us. We should have documentation for this and probably some sort of underlying struct that we set strongly typed values for these on, that we turn `gating_disabled` into.

shell/ci/release/release.sh
jaredallard: We can turn it off by default in the scripts, that's fine. We can make it a global default for _our_ org, that's the main purpose of the box config.

root/checkopslevel.go
artemchernyak-outreach: This is ported from our internal `services` (an internal repo). It is tested there, but happy to duplicate the tests here too. I hate that we have to duplicate this at all, but I also wanted to make this part open.

shell/ci/release/release.sh
artemchernyak-outreach: Oh, I see. For some reason I was stuck on the idea of turning this off at the stencil level.

root/checkopslevel.go
artemchernyak-outreach: I disagree on making this a structure too early. I think we should wait and see what the usage looks like before settling on a structure to use.

jaredallard (approved):

PR #439, opened 2023-02-09
fix: linter not exiting on error

tonymbugua-outreach:
There are instances when running `make test` where an error that is thrown but the terminal will not exit and instead continue. There is a weird issue where if the linter function contains multiple commands and one fails the exit code does not explicitly return an exit value `1`.  This is the work around we came up with for now until we can further figure out a better solution 

## Jira ID

DT-3405

### DIFF
--- a/shell/linters/bash.sh
+++ b/shell/linters/bash.sh
@@ -20,8 +20,8 @@ shellfmt_formatter() {
 }
 
 linter() {
-  run_command "shellcheck" shellcheck_linter || exit 1
-  run_command "shellfmt" shellfmt_linter || exit 1
+  run_command "shellcheck" shellcheck_linter || return 1
+  run_command "shellfmt" shellfmt_linter || return 1
 }
 
 formatter() {
--- a/shell/linters/bash.sh
+++ b/shell/linters/bash.sh
@@ -20,8 +20,8 @@ shellfmt_formatter() {
 }
 
 linter() {
-  run_command "shellcheck" shellcheck_linter
-  run_command "shellfmt" shellfmt_linter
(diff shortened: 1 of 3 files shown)


kaldorn (approved):
`|| exit 1` sets explicit exit code at potential failure points.

jaredallard (commented):
Can we add a doc into `docs` about the linter framework (small amounts of details is fine for now)? I'd like that to reflect why we're doing `exit 1` here right now (working around a bash issue we couldn't quite figure out)

jkinkead:
Calling `exit` from a function rather than returning an error we check seems like a decision that will come back to bite us.

chatchai-outreach:
I think we already handle this in linter https://github.com/getoutreach/devbase/blob/main/shell/linters.sh#L75-L78

jaredallard:
Oh I meant to say this, does `return 1` work instead of `exit`?

tonymbugua-outreach:
> probably do something like linter https://github.com/getoutreach/devbase/blob/main/shell/linters.sh#L75-L78

The problem is that only works if there is one command in the linter function. If there are multiple and the 2 out of the 3rd one fails then no error is going to be returned

tonymbugua-outreach:
> Oh I meant to say this, does `return 1` work instead of `exit`?

I will try and update accordingly if it does

jaredallard:
@tonymbugua-outreach Mind creating a follow up JIRA ticket for the missed docs part here? Just want to make sure we don't forget it!

tonymbugua-outreach:
> @tonymbugua-outreach Mind creating a follow up JIRA ticket for the missed docs part here? Just want to make sure we don't forget it!

oop, forgot to commit it in.

PR #450, opened 2023-02-17
fix(ci): remove docker_layer_caching by default

jaredallard:
This doesn't really add anything for us right now because of our usage of buildx

### DIFF
--- a/orbs/shared/executors/testbed-machine.yml
+++ b/orbs/shared/executors/testbed-machine.yml
@@ -1,7 +1,7 @@
 description: Standard executor for machine runtimes
 machine:
   image: ubuntu-2204:2022.10.2
-  docker_layer_caching: true
+  docker_layer_caching: false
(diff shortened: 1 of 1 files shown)


kaldorn (approved):

orbs/shared/executors/testbed-machine.yml
```diff
@@ -1,7 +1,7 @@
 description: Standard executor for machine runtimes
 machine:
   image: ubuntu-2204:2022.10.2
-  docker_layer_caching: true
+  docker_layer_caching: false
```
jkinkead, line 4: What are you doing to track the impact of this?

CircleCI says we use quite a bit of the layer cache in our services, which means this could cause substantial slowdown.

jaredallard: Taking offline

PR #455, opened 2023-02-22
fix(ci): patch hashicorp APT keyring with older key

jaredallard:

This PR fixes the Hashicorp APT keyring to contain the older key that is still in use in their repositories.

## Jira ID

[ICM-459]

## Notes for your reviewers

[ICM-459]: https://outreach-io.atlassian.net/browse/ICM-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/circleci/machine.sh
+++ b/shell/circleci/machine.sh
@@ -20,7 +20,20 @@ fi
 
 # Add APT repositories we do need.
 if should_install_vault; then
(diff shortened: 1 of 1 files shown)


pbr0ck3r (approved):
Looks good

PR #458, opened 2023-02-25
fix: upgrade stencil modules to remove ruby dep

jaredallard:

The `.tool-versions` is used to calculate all service's dependencies, so things getting the latest commit to stencil-golang won't fully remove `ruby` until this does as well. So, here we are.

## Jira ID

[XX-XX]

### DIFF
--- a/service.yaml
+++ b/service.yaml
@@ -4,65 +4,20 @@ arguments:
   ciOptions:
     skipE2e: true
     skipDocker: true
-  commands: []
   commitGuard: false
-  dependencies:
-    optional: []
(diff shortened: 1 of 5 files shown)


service.yaml
```diff
@@ -4,65 +4,20 @@ arguments:
   ciOptions:
     skipE2e: true
     skipDocker: true
-  commands: []
```
jaredallard, line 7: Cleaned up default values from the bootstrap migration

kaldorn (approved):

PR #465, opened 2023-03-01
feat: speed up calculating dependencies

marekfexa-outreach:
- Currently git clone is performed to read devenv.yaml from repositories (of app or its recursive dependencies)
- This change is about using GH API and getting single file instead

## Jira ID

[QF-841]

## Notes for your reviewers

[QF-841]: https://outreach-io.atlassian.net/browse/QF-841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -289,6 +289,7 @@ func main() { //nolint:funlen,gocyclo // Why: there are no reusable parts to ext
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
 	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
 
 	conf, err := box.EnsureBoxWithOptions(ctx)
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"go/build"
 	"os"
@@ -178,31 +177,6 @@ func parseDevenvConfig(confPath string) (*DevenvConfig, error) {
 	return &dc, nil
 }
 
-// appAlreadyDeployed checks if an application is already deployed, if it is
-// it returns true, otherwise false.
-func appAlreadyDeployed(ctx context.Context, app string) bool {
-	var deployedApps []struct {
-		Name string `json:"name"`
-	}
-
-	b, err := exec.CommandContext(ctx, "devenv", "--skip-update", "apps", "list", "--output", "json").Output()
-	if err != nil {
-		return false
-	}
-
-	if err := json.Unmarshal(b, &deployedApps); err != nil {
-		return false
-	}
-
-	for _, a := range deployedApps {
-		if a.Name == app {
-			return true
-		}
-	}
-
-	return false
-}
-
 // provisionNew destroys and re-provisions a devenv
 func provisionNew(ctx context.Context, deps []string, target string) error { // nolint:unparam // Why: keeping in the interface for now
 	//nolint:errcheck // Why: Best effort remove existing cluster
@@ -346,7 +320,7 @@ func main() { //nolint:funlen,gocyclo // Why: there are no reusable parts to ext
 	// if it's a library we don't need to deploy the application.
 	if dc.Service {
 		log.Info().Msg("Deploying current application into cluster")
-		if osStdInOutErr(exec.CommandContext(ctx, "devenv", "--skip-update", "--with-deps", "apps", "deploy", ".")).Run() != nil {
+		if osStdInOutErr(exec.CommandContext(ctx, "devenv", "--skip-update", "apps", "deploy", "--with-deps", ".")).Run() != nil {
 			log.Fatal().Err(err).Msg("Failed to deploy current application into devenv")
 		}
 	}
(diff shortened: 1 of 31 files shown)


e2e/e2e.go
```diff
@@ -76,18 +75,26 @@ func BuildDependenciesList(ctx context.Context) ([]string, error) {
...
 	possibleFiles := []string{"devenv.yaml", "noncompat-service.yaml", "service.yaml"}
+	gh, err := githubauth.NewClient()
+	if err != nil {
+		return nil, err
+	}
 
 	var dc *DevenvConfig
 	for _, f := range possibleFiles {
-		if _, err := os.Stat(filepath.Join(repoPath, f)); err != nil {
+		config, _, _, err := gh.Repositories.GetContents(ctx, "getoutreach", serviceName, f, nil)
```
jaredallard, line 87: I think this would apply here as well: https://github.com/getoutreach/devenv/pull/494#discussion_r1129633388

e2e/e2e.go
```diff
@@ -127,28 +134,13 @@ func grabDependencies(ctx context.Context, deps map[string]struct{}, serviceName
...
-	cmd := exec.CommandContext(ctx,
-		"git", "clone", "-q", "--depth", "1", "git@github.com:"+path.Join("getoutreach", serviceName), tmpDir,
-	)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	cmd.Stdin = os.Stdin
-	if err := cmd.Run(); err != nil {
-		return errors.Wrapf(err, "failed to clone dependency %s", serviceName)
+	if err != nil {
+		return errors.Wrap(err, "failed to grap dependencies")
```
jaredallard, line 141:
```suggestion
		return errors.Wrap(err, "failed to grab dependencies")
```

e2e/e2e.go
```diff
@@ -76,18 +75,26 @@ func BuildDependenciesList(ctx context.Context) ([]string, error) {
...
+		config, _, _, err := gh.Repositories.GetContents(ctx, "getoutreach", serviceName, f, nil)
+		if err != nil {
 			continue
 		}
-
-		var err error
-		dc, err = parseDevenvConfig(filepath.Join(repoPath, f))
+		content, err := config.GetContent()
 		if err != nil {
+			log.Warn().Str("service", serviceName).Msgf("Unable to get content of file %s", f)
```
jaredallard, line 93: Should we always warn here? I think it's expected that some will fail (e.g., `flagship`) so that seems like extra clutter. Not super attached to this, but I like to log less if it makes sense

e2e/e2e.go
```diff
@@ -76,18 +75,26 @@ func BuildDependenciesList(ctx context.Context) ([]string, error) {
...
+			continue
+		}
+		if err := yaml.NewDecoder(strings.NewReader(content)).Decode(&dc); err != nil {
+			log.Warn().Str("service", serviceName).Msgf("Unable to parse %s", f)
```
jaredallard, line 97: Instead of `%s` should we use a field for the filename? (applies to both)

e2e/e2e.go
marekfexa-outreach: yeah, I wanted to keep it there only because it always was there, but I hate it TBH

jaredallard: Maybe we can just throw it being `log.Debug()` for now and call it good? We should have a `--debug` flag that works

marekfexa-outreach: 👍  Added

e2e/e2e.go
```diff
@@ -74,21 +74,41 @@ func BuildDependenciesList(ctx context.Context) ([]string, error) {
...
 }
 
+// getConfig gets config file from GitHub and parses it into DevenvConfig
+func getConfig(ctx context.Context, serviceName string, gh *github.Client, configFileName string) (*DevenvConfig, error) {
+	r, _, err := gh.Repositories.DownloadContents(ctx, "getoutreach", serviceName, configFileName, nil)
+	l := log.With().Str("service", serviceName).Str("file", configFileName).Logger()
+	defer func() {
+		if err := r.Close(); err != nil {
+			l.Warn().Msg("Unable to close GH reader")
+		}
```
jaredallard, line 84: nit: Generally, we can throw this away; I know it's not the most intuitive thing in the world to ignore that...

e2e/e2e.go
```diff
@@ -74,21 +74,41 @@ func BuildDependenciesList(ctx context.Context) ([]string, error) {
...
+	}()
+	if err != nil {
+		l.Debug().Msg("Unable to find file in GH")
+		return nil, err
+	}
+	var dc *DevenvConfig
```
jaredallard, line 90: nit: Generally only decode into the non-pointer type and reference it in the return, it's entirely a style nitpick.

e2e/e2e.go
```diff
@@ -74,21 +74,41 @@ func BuildDependenciesList(ctx context.Context) ([]string, error) {
 	return depsList, nil
 }
 
+// getConfig gets config file from GitHub and parses it into DevenvConfig
+func getConfig(ctx context.Context, serviceName string, gh *github.Client, configFileName string) (*DevenvConfig, error) {
+	r, _, err := gh.Repositories.DownloadContents(ctx, "getoutreach", serviceName, configFileName, nil)
```
jaredallard, line 79: Change `getoutreach` to be the org from the box config?

marekfexa-outreach: Fixed, yeah, I was too lazy

e2e/e2e.go
marekfexa-outreach: Fixed

e2e/e2e.go
marekfexa-outreach: Fixed

jaredallard (approved):

PR #477, opened 2023-03-10
fix: handle NPE and revise error messages

chatchai-outreach:

- handle NPE
- revise error messages and add log for readability

** no service entry found in opslevel **
Screenshot 2023-03-10 at 12 37 05 PM

** there's an entry but no lifecycle is assigned in opslevel **
Screenshot 2023-03-10 at 1 01 09 PM

** service maturity does not meet the requirement **
Screenshot 2023-03-10 at 1 10 12 PM

## Jira ID

**requirement**: https://outreach-io.atlassian.net/browse/DT-2954
**bug fix**: https://outreach-io.atlassian.net/browse/DT-3563

### DIFF
--- a/root/checkopslevel.go
+++ b/root/checkopslevel.go
@@ -5,42 +5,69 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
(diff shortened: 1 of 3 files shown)


jaredallard (approved):
Looks legit!

PR #480, opened 2023-03-14
feat: Adding global ASDF plugin ability (+golang) while ensuring asdf

deregtd:


shell/lib/asdf.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 # Utilities for working with asdf
 
+asdf_global_plugins=("golang 1.19.5")
```
jaredallard, line 4: Should we read the current go version out of the projects `.tool-versions` instead of hardcoding this here?

deregtd: I thought about that, but I'm torn -- that ties the version we expect to use to install plugins to the version that the repo is using, which feels like we should keep it dissociated.  Then you have your build system's build system (to assemble your build system), then separately your build system builds your app, and they're uncoupled.

deregtd: Talked offline and updated the method

shell/lib/asdf.sh
```diff
@@ -1,6 +1,15 @@
...
 # Utilities for working with asdf
 
+# asdf_global_plugins is a list of global plugins to always make sure exist, usually
+# to ensure that /other/ asdf plugins have a valid build environment to build
+# themselves with.  This script will attempt to _first_ install everything from this
+# global plugins list before going through the tool-versions from the local app.
+# Any global plugin that has a matching version from inside the app will set that
+# version globally.  If the app does not have the global plugin in its local
+# tool-versions file, it will fall back to the hardcoded version in here.
+asdf_global_plugins=("golang 1.19.5")
```
jaredallard, line 11: Can the trade-off for this "global" version be that we read it from the `devbase` `.tool-versions` file? That's what we've been using, more or less, for global versions and would be easily updatable then.

deregtd:
closing this PR in favor of a more extensive work item to do this more comprehensively/properly

PR #481, opened 2023-03-14
fix(generics): bump to golangci-lint version

AndrewWinterman:

updates golangci-lint version one minor version to enable generics support.

## Jira ID

[QSS-18000]


jaredallard (changes requested):
Can't because of https://github.com/golangci/golangci-lint/issues/3086 :(

AndrewWinterman:
oh deer https://github.com/ent/ent/pull/3360-- they've made a fix in ent for this, and golanglint-ci will be updated very shortly to fix it, but probably not in time for my project

PR #485, opened 2023-03-18
fix: Install gh for arm

voslartomas:

Conditionally installs amd64 or arm64 gh cli instead of always amd64

## Jira ID

[XX-XX]


deregtd (approved):
Conditionally installs amd64 or arm64 gh cli instead of always amd64

PR #490, opened 2023-03-22
fix: send notification if opslevel check fails

chatchai-outreach:

we want to get notified if releases fails because of opslevel check

## Jira ID

[XX-XX]


tonymbugua-outreach (approved):
LGTM - notify if it fails

PR #494, opened 2023-03-23
feat: Add `SHUFFLE=disabled` test env option

richardlarocque:
Adds the option to remove the `-shuffle=on` flag.  The use of `-shuffle`
disables test caching, which is not a great experience for TDD.  Many
devs will want to turn this off during local development.

### DIFF
--- a/shell/test.sh
+++ b/shell/test.sh
@@ -79,6 +79,13 @@ GO_TEST_TIMEOUT="${GO_TEST_TIMEOUT:-}"
 # will not be passed to 'go test'. Defaults to 'enabled'.
 RACE="${RACE:-enabled}"
 
+# SHUFFFLE determines if 'go test -shuffle' should be enabled or not. If set to
+# 'disabled', test order randomization will not be enabled and the flag will
+# not be passed to 'go test'. Note that `-shuffle` is incompatible with test
+# caching, so disabling it can result in a substantial speedup in development.
+# Defaults to 'enabled'.
+SHUFFLE="${SHUFFLE:-enabled}"
+
 # TEST_OUTPUT_FORMAT is the format to pass to gotestsum. If not set,
 # the default value of "dots-v2" will be used. If CI is set, the default
 # value is "pkgname". If 'TEST_FLAGS' contains '-v', the default value
@@ -110,7 +117,7 @@ REPODIR=$(get_repo_directory)
 # Catches test dependencies by shuffling tests if the installed Go version supports it
 currentver="$(go version | awk '{ print $3 }' | sed 's|go||')"
 requiredver="1.17.0"
-if [[ "$(printf '%s\n' "$requiredver" "$currentver" | sort -V | head -n1)" == "$requiredver" ]]; then
+if [[ "$(printf '%s\n' "$requiredver" "$currentver" | sort -V | head -n1)" == "$requiredver" && $SHUFFLE != "disabled" ]]; then
   TEST_FLAGS+=(-shuffle=on)
 fi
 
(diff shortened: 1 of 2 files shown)


richardlarocque:
I'm opening this PR as a conversation-starter.  I'm very much open to feedback on other approaches.

I think one could argue that `-shuffle=on` should be on only In CI, not during local development.  The loss of test caching is a pretty big hit to a developer's edit-test-fix iteration loop, and this `SHUFFLE=disabled` flag is not very discoverable.

jaredallard:
> > I'm worried about how we'd expose this to users. If we turn it off by default, outside of CI, we could hit issues where CI seems unreproducible because of a test dependency unknown to the user.
> 
> Just brainstorming here, but could we emit a log like:
> 
> ```
> The randomly-generated seed for this test is 1234523.  To reproduce this test run as closely as possible:
> 
> $ SEED=1234523 CI_MODE=true make test
> ```

It looks like, when a test fails, it emits the seed being used ([example [link]](https://app.circleci.com/pipelines/github/getoutreach/httpx/529/workflows/a214dfa5-fb7d-44c9-8a21-3c463138f068/jobs/1137?invite=true#step-106-110)):

```
✖  pkg/handlers (153ms) (coverage: 18.8% of statements in ./...) (-test.shuffle 1681144907598172136)
✓  pkg/proxy (60ms) (coverage: 14.5% of statements in ./...)
✖  pkg/fetch (1m0.161s) (coverage: 47.3% of statements in ./...) (-test.shuffle 1681144907948383367)
```

richardlarocque:
@jaredallard given that we have the `-test.shuffle` instructions being printed in the case of failure in Circle, and someone reasonably determined could use that to reproduce the Circle test ordering locally, would it be OK to proceed with this change?

This is coming back to top-of-mind because I'm trying to get the ORCS team and copycat devs in particular to make better use of e2e tests.  I have a hack script that can iterate on e2e tests in ~2s:
```
time scripts/quick-e2e.sh ./...
...
real    0m1.892s
user    0m6.638s
sys     0m1.697s
```

while `make test-e2e` takes much longer due to shuffling:
```
time make test-e2e
...
real    0m43.531s
user    1m42.482s
sys     0m19.413s
```

jaredallard:
@richardlarocque Sorry for the delay. I think proceeding with this change is fine! Sorry for the mess I made of rewriting the `test.sh` script...

richardlarocque:
Thanks!  I'll expect a challenging rebase followed by a quick approval, then.  :)

(Probably not today, though. Today is full of meetings, so I won't get to it.)

richardlarocque:
@jaredallard this should be rebased and ready for review now.

shell/test.sh
```diff
@@ -79,6 +79,13 @@ GO_TEST_TIMEOUT="${GO_TEST_TIMEOUT:-}"
 # will not be passed to 'go test'. Defaults to 'enabled'.
 RACE="${RACE:-enabled}"
 
+# SHUFFFLE determines if 'go test -shuffle' should be enabled or not. If set to
```
jaredallard, line 82: We're word wrapping at 72 here, but I think 80(?) is fine as well :D

shell/test.sh
```diff
@@ -79,6 +79,13 @@ GO_TEST_TIMEOUT="${GO_TEST_TIMEOUT:-}"
...
+# 'disabled', test order randomization will not be enabled and the flag will
+# not be passed to 'go test'. Note that `-shuffle` is incompatible with test
+# caching, so disabling it can result in a substantial speedup in development.
+# Defaults to 'enabled'.
```
jaredallard, line 86:
```suggestion
# Defaults to 'enabled' if the version of Go supports it (>=1.17).
```

jaredallard (approved):
Thanks for the contribution!

PR #497, opened 2023-03-28
feat: build docker early to speedup e2e

marekfexa-outreach:
- Docker build takes e.g. in `rms` 6 minutes which were previously added to `make e2e` execution time
- This change is about not building docker in the critical path, but together with app dependencies deployment where CPU load is low

## Jira ID

[QF-848]

## Notes for your reviewers

[QF-848]: https://outreach-io.atlassian.net/browse/QF-848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -206,7 +206,7 @@ func appAlreadyDeployed(ctx context.Context, app string) bool {
 }
 
 // provisionNew destroys and re-provisions a devenv
-func provisionNew(ctx context.Context, deps []string, target string) error { // nolint:unparam // Why: keeping in the interface for now
+func provisionNew(ctx context.Context, target string) error { // nolint:unparam // Why: keeping in the interface for now
 	//nolint:errcheck // Why: Best effort remove existing cluster
 	exec.CommandContext(ctx, "devenv", "--skip-update", "destroy").Run()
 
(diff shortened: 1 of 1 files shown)


e2e/e2e.go
```diff
@@ -343,10 +349,26 @@ func main() { //nolint:funlen,gocyclo // Why: there are no reusable parts to ext
...
 			}
+
+			wg.Add(1)
+
+			// Build docker sooner and out of critical path to speed things up.
+			// Docker build in devenv apps deploy . will be superfast then.
+			go func(wg *sync.WaitGroup) {
+				defer wg.Done()
+				log.Info().Msg("Starting early docker build")
+				if err := exec.CommandContext(ctx, "make", "docker-build").Run(); err != nil {
```
jaredallard, line 364: If this fails, we're going to have no output. I imagine the idea is that the later `docker apps deploy .` will show the error, right?

Or should we just exit early if this fails and dump the entire output?

e2e/e2e.go
```diff
@@ -359,6 +381,8 @@ func main() { //nolint:funlen,gocyclo // Why: there are no reusable parts to ext
 		log.Fatal().Err(err).Msg("Failed to parse devenv.yaml, cannot run e2e tests for this repo")
 	}
 
+	wg.Wait() // To ensure that docker build is finished
```
jaredallard, line 384: Should we log that we're waiting on the docker build here? If we're showing no output it could appear that it's hanging

e2e/e2e.go
marekfexa-outreach: My thinking here is that we don't care much here about the output, since docker build will be called again later in `devenv apps deploy .` (it will be much faster since images were already built). So we will see output anyway from there.

jaredallard (approved):

PR #500, opened 2023-03-28
feat: Deploy apps in parallel to speedup e2e

marekfexa-outreach:
- Run all `devenv apps deploy` of current app dependencies in parallel to speed up `make e2e` 

## Jira ID

[XX-XX]


marekfexa-outreach:
This has to be done on devenv side, using --with-deps and supporting parallelization there

PR #502, opened 2023-04-04
fix(asdf): dont automatically install amd64 versions

jaredallard:

Previously, we'd always install amd64 versions of binaries if `asdf install` failed. This was our only way to determine if a binary was not compatible with `darwin/arm64`. However, since then, we've moved away from anything needing amd64 compat, and the few that do we install with `orc setup`. As such, I've removed the automatic handling of this and replaced it with a note.

We've had a few users have the step fail due to network flakiness, or other issues, and get the amd64 version without knowing. This is problematic for many reasons, so that's the main motivation of this change.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/lib/asdf.sh
+++ b/shell/lib/asdf.sh
@@ -128,13 +128,26 @@ asdf_install_retry() {
   local plugin="$1"
   local version="$2"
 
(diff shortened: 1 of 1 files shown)


kaldorn (approved):
That seems reasonable.

PR #506, opened 2023-04-05
fix: re-stencil

tonymbugua-outreach:
restencil with channel: rc

## Jira ID

[XX-XX]


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.10.0
+  shared: getoutreach/shared@dev:local
```
jaredallard, line 5: I think you'll need to revert this line, it's a nuance of devbase replacing itself. Will need to add some logic to stencil-circleci to use some other version for this. Dunno yet.

package.json
```diff
@@ -1,7 +1,6 @@
 {
   "license": "UNLICENSED",
   "devDependencies": {
-    "@outreach/semantic-release-circleci-orb": "^1.1.9",
```
jaredallard, line 4: This needs to be added back for now, eventually @kaldorn and I had plans to fix this.

Should re-run `yarn` after adding back.

PR #507, opened 2023-04-06
feat(golangci-lint): 1.52.2

jaredallard:

Upgrades `golangci-lint` to get more linter fixes. This needs to be tested on a repo using `ent` before it can be merged to make sure that that isn't still broken.

## Jira ID

[XX-XX]


jaredallard:
Nope, blocked on https://github.com/golangci/golangci-lint/issues/3711

PR #510, opened 2023-04-11
fix(debug): always use dlv debug, better docs

jaredallard:

This PR moves us from `dlv exec` for in container debugger runs to `dlv debug`. The main goal of this change is to prevent needing to compile binaries with GC flags that only `dlv` knows of.

I also focused on adding some docs for this script and some more touch points for being able to debug different packages, since it's useful for developers to be able to set these.

## Jira ID

[DT-3662]

## Notes for your reviewers

[DT-3662]: https://outreach-io.atlassian.net/browse/DT-3662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/debug.sh
+++ b/shell/debug.sh
@@ -43,6 +43,10 @@ HEADLESS="${HEADLESS:-$IN_CONTAINER}"
 # Why: Declared for documentation/discoverability purposes.
 # shellcheck disable=SC2269
 DLV_PORT="${DLV_PORT}"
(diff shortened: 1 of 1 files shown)


jaredallard:
FYI @clevelittlefield-outreach

clevelittlefield-outreach:
Yay!

chatchai-outreach (approved):

PR #515, opened 2023-04-18
feat: move circleci e2e tests to devbase

kaldorn:
Moves the e2e testing into devbase so that e2e run in CircleCI is the same as e2e run locally.

## Jira ID

[DT-2172]

## Notes for your reviewers

[DT-2172]: https://outreach-io.atlassian.net/browse/DT-2172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonymbugua-outreach (approved):

jaredallard:
Did we do any testing of this? I'd like to be sure this is working before we switch over to `devenv apps e2e` everywhere

kaldorn:
I tested, but it's been a while. IIRC I need to not remove those environmental variable declarations, and the devenv needs to be spun up.

kaldorn:
As this PR sits, it tests fine against copycat.

jaredallard (approved):

malept (changes requested):
I have some concerns, namely that in CircleCI, if you have failing tests, it doesn't appear to upload test results properly to show exactly which tests failed (it can't find the test results file).

jaredallard (commented):
Upgrade gobox and I think we should be good giving this a shot (if we're still confident it works)

PR #520, opened 2023-04-20
feat:  pre-setup-script for tests

tonymbugua-outreach:
Users would like the ability to add packages or even run a script before running unit tests. This PR will allow users to specify a path that contains a shell script to run before test.

Tested successfully with this [CircleCI pipeline](https://app.circleci.com/pipelines/github/getoutreach/devbase/2701/workflows/a19e9c98-9ae5-44de-b050-acbca4f1d6c4/jobs/6509)

## Jira ID

DT-3641

### DIFF
--- a/orbs/shared/jobs/save_cache.yaml
+++ b/orbs/shared/jobs/save_cache.yaml
@@ -7,6 +7,9 @@ parameters:
     default: _json_key
     description: Username to use when fetching images from a registry
     type: string
+  pre_test_script:
+    description: Environment variable that contains a shell path to run before running tests in CircleCI
+    type: string
   docker_password:
     default: $GCLOUD_SERVICE_ACCOUNT
     description: Environment variable to read docker password from
--- a/orbs/shared/jobs/test.yaml
+++ b/orbs/shared/jobs/test.yaml
@@ -8,7 +8,7 @@ parameters:
     description: Username to use when fetching images from a registry
     type: string
   pre_setup_script:
-    description: Environment variable that contains a shell path to execute before running tests in CircleCI
+    description: Environment variable that contains an executable to run before running tests in CircleCI. This may not include arguments.
     type: string
   docker_password:
     default: $GCLOUD_SERVICE_ACCOUNT
--- a/orbs/shared/jobs/test.yaml
+++ b/orbs/shared/jobs/test.yaml
@@ -7,6 +7,9 @@ parameters:
     default: _json_key
     description: Username to use when fetching images from a registry
     type: string
+  pre_setup_script:
+    description: Environment variable that contains a shell path to execute before running tests in CircleCI
+    type: string
   docker_password:
     default: $GCLOUD_SERVICE_ACCOUNT
     description: Environment variable to read docker password from
@@ -41,7 +44,7 @@ executor:
 
 environment:
   AWS_REGION: << parameters.aws_region >>
-
+  PRE_SETUP_SCRIPT: << parameters.pre_setup_script >>
 steps:
   - when:
       condition:
--- a/orbs/shared/jobs/test.yaml
+++ b/orbs/shared/jobs/test.yaml
@@ -8,7 +8,7 @@ parameters:
     description: Username to use when fetching images from a registry
     type: string
   pre_setup_script:
-    description: Environment variable that contains an executable to run before running tests in CircleCI. This may not include arguments.
+    description: If set, the executable to run before running tests in CircleCI. This may not include arguments.
     type: string
   docker_password:
     default: $GCLOUD_SERVICE_ACCOUNT
--- a/shell/circleci/setup.sh
+++ b/shell/circleci/setup.sh
@@ -35,7 +35,7 @@ fi
 if [[ -n $PRE_SETUP_SCRIPT ]]; then
   echo "⚙️ Running setup script \"${PRE_SETUP_SCRIPT}\" (from PRE_SETUP_SCRIPT)"
   # shellcheck source=/dev/null
-  source "${PRE_SETUP_SCRIPT}"
+  "${PRE_SETUP_SCRIPT}"
 fi
 
 # Setup a box stub
--- a/shell/circleci/setup.sh
+++ b/shell/circleci/setup.sh
@@ -31,8 +31,11 @@ if [[ -n $TEST_RESULTS ]]; then
   mkdir -p "$TEST_RESULTS"
 fi
 
-if [[ -n $PRETEST_SCRIPT ]]; then
-  source "${PRETEST_SCRIPT}
+# run prescript if user specified to install packages etc. before tests
+if [[ -n $PRE_SETUP_SCRIPT ]]; then
+  echo "⚙️ Running setup script \"${PRE_SETUP_SCRIPT}\" (from PRE_SETUP_SCRIPT)"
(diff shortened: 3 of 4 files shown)


shell/circleci/setup.sh
```diff
@@ -31,6 +31,13 @@ if [[ -n $TEST_RESULTS ]]; then
   mkdir -p "$TEST_RESULTS"
 fi
 
+# run prescript if user specified to install packages etc. before tests
+if [[ -n $PRETEST_SCRIPT ]]; then
+  echo "⚙️ running prescript file ${PRETEST_SCRIPT}"
```
jaredallard, line 36: nit: match case of other statements

```suggestion
  echo "⚙️ Running setup script \"${PRETEST_SCRIPT}\" (from PRETEST_SCRIPT)"
```

orbs/shared/jobs/test.yaml
```diff
@@ -7,6 +7,9 @@ parameters:
     default: _json_key
     description: Username to use when fetching images from a registry
     type: string
+  pre_test_script:
```
jaredallard, line 10: Given this is ran by the `setup.sh` script, it might make more sense to call this `pre_setup_script`, since this can be run in other contexts than just the `test.sh` script.

tonymbugua-outreach: fixed

shell/circleci/setup.sh
```diff
@@ -31,6 +31,13 @@ if [[ -n $TEST_RESULTS ]]; then
   mkdir -p "$TEST_RESULTS"
 fi
 
+# run prescript if user specified to install packages etc. before tests
+if [[ -n $PRE_SETUP_SCRIPT ]]; then
```
jkinkead, line 35: This should also check if this is a file (or executable; see below), and error if it's not:
```suggestion
if [[ -n $PRE_SETUP_SCRIPT ]]; then
  if [[ ! -x $PRE_SETUP_SCRIPT ]]; then
    echo "PRE_SETUP_SCRIPT was not an executable; got "$PRE_SETUP_SCRIPT'"
    exit 1
  fi
```

shell/circleci/setup.sh
```diff
@@ -31,6 +31,13 @@ if [[ -n $TEST_RESULTS ]]; then
...
+  echo "⚙️ Running setup script \"${PRE_SETUP_SCRIPT}\" (from PRE_SETUP_SCRIPT)"
+  # shellcheck source=/dev/null
+  source "${PRE_SETUP_SCRIPT}"
```
jkinkead, line 38: Is there a reason we are sourcing instead of executing, e.g. doing:
```suggestion
  "${PRE_SETUP_SCRIPT}"
```
?

Sourcing means:
* this has to be a bash script
* shell variables will be inherited in the current context

shell/circleci/setup.sh
tonymbugua-outreach: decided this is not needed

shell/circleci/setup.sh
tonymbugua-outreach: decided this is needed. We want users to be able to run it against current shell

orbs/shared/jobs/test.yaml
```diff
@@ -7,6 +7,9 @@ parameters:
     default: _json_key
     description: Username to use when fetching images from a registry
     type: string
+  pre_setup_script:
+    description: Environment variable that contains a shell path to execute before running tests in CircleCI
```
jkinkead, line 11: Docs tweaks:
```suggestion
    description: Environment variable that contains an executable to run before running tests in CircleCI. This may not include arguments.
```

shell/circleci/setup.sh
tonymbugua-outreach: Removed source

orbs/shared/jobs/save_cache.yaml
```diff
@@ -7,6 +7,9 @@ parameters:
     default: _json_key
     description: Username to use when fetching images from a registry
     type: string
+  pre_test_script:
+    description: Environment variable that contains a shell path to run before running tests in CircleCI
```
jkinkead, line 11: This needs to be removed - right? You didn't change the save cache script from what I can tell ...

orbs/shared/jobs/test.yaml
```diff
@@ -7,6 +7,9 @@ parameters:
     default: _json_key
     description: Username to use when fetching images from a registry
     type: string
+  pre_setup_script:
+    description: Environment variable that contains an executable to run before running tests in CircleCI. This may not include arguments.
```
jkinkead, line 11: This is not an environment variable, it's just an argument:
```suggestion
    description: If set, the executable to run before running tests in CircleCI. This may not include arguments.
```

shell/circleci/setup.sh
```diff
@@ -31,6 +31,13 @@ if [[ -n $TEST_RESULTS ]]; then
   mkdir -p "$TEST_RESULTS"
 fi
 
+# run prescript if user specified to install packages etc. before tests
+if [[ -n $PRE_SETUP_SCRIPT ]]; then
+  echo "⚙️ Running setup script \"${PRE_SETUP_SCRIPT}\" (from PRE_SETUP_SCRIPT)"
```
jkinkead, line 36: Users set a lower-case argument, right?
```suggestion
  echo "⚙️ Running setup script \"${PRE_SETUP_SCRIPT}\" (from pre_setup_script)"
```

jkinkead (commented):
How have you been testing this, out of curiosity?

orbs/shared/jobs/save_cache.yaml
tonymbugua-outreach: This is needed for the `save_cache` job because the test object is gatting passed in on this [line](https://github.com/getoutreach/devbase/blob/main/.circleci/config.yml#L80)

tonymbugua-outreach:
> How have you been testing this, out of curiosity?

Yes I was testing with this [pipeline]. 
 had a basic echo `SUCCESSFULLY RUNNING FROM PRESCRIPT
` print during the "Setup CI Environment" job(https://app.circleci.com/pipelines/github/getoutreach/devbase/2701/workflows/a19e9c98-9ae5-44de-b050-acbca4f1d6c4/jobs/6509)

orbs/shared/jobs/save_cache.yaml
jkinkead: Huh. I clearly need to read more docs here to understand what's going on. :)

Can you update this description to match the other one?

jkinkead: Huh. I clearly need to read more docs here to understand what's going on. :)

Can you update this description to match the other one?

If the tests are being invoked, do you also need an `environment` block that sets the environment variable?

jkinkead (approved):
LGTM with one request and one question.

orbs/shared/jobs/save_cache.yaml
tonymbugua-outreach: Yes. There is an env block that is set in the test resource defined [here](https://github.com/getoutreach/devbase/pull/520/files#diff-b493fedbc0e0080b57980c59870e378f2a69d198e9f6b3ccdd4a638dc05f5e03R47)

PR #521, opened 2023-04-21
fix: opt-out devconfig early

marekfexa-outreach:
- Allow opt-out from running devconfig in parallel with apps deploy (introduced in https://github.com/getoutreach/devbase/pull/499)
- Based on feedback from copilot project, which has devconfig override that required application to be deployed prior to running it https://github.com/getoutreach/copilot/blob/main/scripts/devconfig.override.sh
- Opt out is based on REQUIRE_DEVCONFIG_AFTER_DEPLOY env var, which can be set on project level in CircleCI 

## Jira ID

[XX-XX]


guoweis-outreach (approved):

kaldorn (approved):
This looks reasonable to me. Approving.

PR #522, opened 2023-04-21
feat: bump golangci to 1.52.2

kaldorn:
Bumping golangci because we're experiencing this type of issue: https://github.com/golangci/golangci-lint/issues/3582

Unfortunately this looks like `ent` is still broken.

## Jira ID
OCE WORK
[XX-XX]

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -15,7 +15,7 @@ nodejs 18.14.1
 ## <<Stencil::Block(toolver)>>
 # The below tools are used by all repositories when using devbase
 mage 1.14.0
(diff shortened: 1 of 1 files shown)


jkinkead (approved):

jkinkead (changes requested):
I think we decided this was Bad, Actually, because it breaks existing unit tests.

We should instead explore ways for customers to set these tool versions themselves.

jaredallard:
Cloing in favor of something like https://github.com/getoutreach/devbase/pull/523 (w/ the considerations in my comment)

PR #523, opened 2023-04-22
feat: allow repo level .tool-version overrides

kaldorn:
Allows a repo to have a `.tool-versions` for golangci-lint, specificaly. This is only needed until this issue is fixed: https://github.com/golangci/golangci-lint/issues/3711

## Jira ID

[XX-XX]

### DIFF
--- a/shell/lib/asdf.sh
+++ b/shell/lib/asdf.sh
@@ -40,9 +40,20 @@ asdf_devbase_exec() {
   # shellcheck disable=SC2155
   local tool_env_var="$(echo "$tool" | tr '[:lower:]-' '[:upper:]_')"
 
-  # Why: We're OK with this being the way it is.
-  # shellcheck disable=SC2155
-  local version="$(asdf_get_version_from_devbase "$tool")"
+  # Check if there's a repo level .tool-versions and if so, allow that to override the devbase
+  # version of a specific tool
+  if [[ $tool == "golangci-lint" ]] && grep "golangci-lint" .tool-versions; then
+    # Why: We're OK with this being the way it is.
+    # shellcheck disable=SC2155
+    local version=$(grep golangci-lint <.tool-versions | awk '{print $2}')
+    echo "found repo override .tool-versions of golangci-lint, using $version instead"
+    echo "$version"
+  else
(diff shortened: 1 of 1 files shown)


shell/lib/asdf.sh
```diff
@@ -40,9 +40,19 @@ asdf_devbase_exec() {
   # shellcheck disable=SC2155
   local tool_env_var="$(echo "$tool" | tr '[:lower:]-' '[:upper:]_')"
 
-  # Why: We're OK with this being the way it is.
-  # shellcheck disable=SC2155
-  local version="$(asdf_get_version_from_devbase "$tool")"
+  # Check if there's a repo level .tool-versions and if so, allow that to override the devbase
+  # version of a specific tool
+  if [[ $tool == "golangci-lint" ]] && grep "golangci-lint" .tool-versions; then
```
jkinkead, line 45: Can we just use an environment variable instead? This is going to have side-effects in repos, potentially, and having a one-off environment variable would be better.

Will this handle installing the right version if it hasn't already been installed?

shell/lib/asdf.sh
```diff
@@ -40,9 +40,24 @@ asdf_devbase_exec() {
   # shellcheck disable=SC2155
   local tool_env_var="$(echo "$tool" | tr '[:lower:]-' '[:upper:]_')"
 
-  # Why: We're OK with this being the way it is.
-  # shellcheck disable=SC2155
-  local version="$(asdf_get_version_from_devbase "$tool")"
+  # Allow overriding the golangci-lint version with an environmental variable
+  if [[ $tool = "golangci-lint" ]] && [[ "${GOLANGCI_LINT_VER}" ]]; then
```
jaredallard, line 44: Putting a version in the top block in the `.tool-versions` should already be overriding... so I'm not really sure this is required.

jaredallard: Talking more offline -- realized that this _is_ required for devbase managed versions. We need the ability for the `.tool-versions` in the root of the repo to override the devbase specified version, which isn't a thing today.

jaredallard:
Replacing with https://github.com/getoutreach/devbase/pull/536

PR #534, opened 2023-05-19
fix: increase golangci-lint timeout to 20m to unblock clerkcommons

nissim-natanov-outreach:
The golangci-lint tool needs more time in clerkcommons repo, increasing the timeout to 20m to unblock the repo. Trying to upgrade to latest golangci-lint version did not help, it still times out after 10m.

## Jira ID

[XX-XX]


jaredallard (changes requested):
Globally bumping the timeout to help one repository is a recipe for bubbling CI cost. Blocking unless this is shown to the the absolute only option.

Tomorrow we have functionality to enable replacing the golangci-lint version coming out that will solve this for the long-term.

nissim-natanov-outreach:
This PR is now referenced directly by clerkcommons to unblock its pipeline: https://github.com/getoutreach/clerkcommons/blob/37c296c144e249942118224f4d30195ea0086197/stencil.lock#L5

Please do not close it while this reference exists or else CI will be blocked!

@jaredallard Jared, replacing the golangci-cli version isn't enough - even latest version still runs > 10m, we may need it to push further as the amount of auto-gen code grows. Maybe we can somehow tell the tool to skip auto-gen code ...

Either way, we need to either be able to disable linter, or increase its timeout or let it skip the auto-gen files.

nissim-natanov-outreach:
seems like the linter is now below 6m, closing for now (I removed the use of the branch from clerkcommons)

PR #538, opened 2023-05-22
fix(setup-devenv): use new gcp kubectl key

jaredallard:

It looks like they changed their GPG key path and then rotated their keys and didn't update the old path (but still serve it). So, we're using the new path. Updated to their new instructions, including a new path to download it to. Love it.

## Jira ID

[XX-XX]


jkinkead (approved):
Fix makes me feel :), cause makes me feel :(

PR #541, opened 2023-05-23
fix(orb/job/test): dont require pre_setup_script

jaredallard:
Fixes the `pre_setup_script` argument to not be required. As per the
[CircleCI documentation](https://circleci.com/docs/reusing-config/#parameter-syntax:~:text=N/A-,default,-The%20default%20value) an argument is required if there is not
'default' key set.

### DIFF
--- a/orbs/shared/jobs/test.yaml
+++ b/orbs/shared/jobs/test.yaml
@@ -8,6 +8,7 @@ parameters:
     description: Username to use when fetching images from a registry
     type: string
   pre_setup_script:
+    default: ""
     description: If set, the executable to run before running tests in CircleCI. This may not include arguments.
     type: string
   docker_password:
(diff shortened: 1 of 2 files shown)


orbs/shared/jobs/test.yaml
```diff
@@ -8,6 +8,7 @@ parameters:
     description: Username to use when fetching images from a registry
     type: string
   pre_setup_script:
+    default: ""
```
jkinkead, line 11: `save_cache` also needs updating, I believe. I has this added as well.

tonymbugua-outreach: added it to the save_cache file

tonymbugua-outreach (approved):

PR #548, opened 2023-05-24
feat: support rtx instead of asdf

jaredallard:
This PR adds support for [rtx](https://github.com/jdxcode/rtx) instead of the asdf version manager.
This functionality can be disabled with `DEVBASE_RTX_DISABLE_RTX` and is
enabled if `rtx` exists in `PATH`.

We're mostly shimming `rtx` support because it is compatible out of the
box with `asdf`. Longer term we should have hot swappable
implementations using interfaces when we move this functionality to Go.
The main driver of this is a desire to move away from `asdf` longer
term.

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -12,7 +12,7 @@ APP                 ?= unset
 ORG                 ?= getoutreach
 
 # Transition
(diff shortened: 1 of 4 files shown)


jaredallard:
This _works_, but it's pretty messy. I'm going to avoid `rtx` support until we do the work to move the VM handling into Go.

PR #549, opened 2023-05-25
fix: allow ctrl-c to kill headless delve in make debug

joeysmithoutreach:

Allow users to kill make debug with ctrl from inside the devenv

## Jira ID

[DT-3698](https://outreach-io.atlassian.net/browse/DT-3698)

## Notes for your reviewers

[DT-3698]: https://outreach-io.atlassian.net/browse/DT-3698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/debug.sh
+++ b/shell/debug.sh
@@ -106,13 +106,12 @@ else
     # a container so the output can be processed by devspace.
     echo -e "\n\n\n\n\n\n\n\n" >>"$DEV_CONTAINER_LOGFILE"
     "${delve[@]}" >>"$DEV_CONTAINER_LOGFILE" 2>&1 &
-    IS_LOGGING="true"
   fi
 
   DLV_PID=$!
   echo "delve pid is: $DLV_PID"
 
-  if [[ $IS_LOGGING == "true" ]]; then
+  if [[ $HEADLESS != "false" ]]; then
     # tail to watch logs here
     tail -n 5 -f "$DEV_CONTAINER_LOGFILE"
   fi
--- a/shell/debug.sh
+++ b/shell/debug.sh
@@ -111,7 +111,7 @@ else
   DLV_PID=$!
   echo "delve pid is: $DLV_PID"
 
-  if [[ $HEADLESS != "false" ]]; then
+  if [[ $IN_CONTAINER != "false" ]]; then
     # tail to watch logs here
     tail -n 5 -f "$DEV_CONTAINER_LOGFILE"
   fi
--- a/shell/debug.sh
+++ b/shell/debug.sh
@@ -83,11 +83,36 @@ if [[ $HEADLESS == "true" ]]; then
   )
 fi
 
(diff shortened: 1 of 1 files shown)


jaredallard (commented):
Thinking about this, do we need to ensure we only do this in headless mode? I'm worried that this will break using the debugger outside headless mode right now.

joeysmithoutreach:
@jaredallard If I understand correctly, `HEADLESS="${HEADLESS:-$IN_CONTAINER}"` in line 39 means that HEADLESS is always true if you're in a container.

jaredallard:
> @jaredallard If I understand correctly, `HEADLESS="${HEADLESS:-$IN_CONTAINER}"` in line 39 means that HEADLESS is always true if you're in a container.

It means `HEADLESS` defaults to true when `IN_CONTAINER`, if not set. If someone sets the env var `HEADLESS`, they will have their debugger output piped to the file.

I don't think that your change actually breaks this, in retrospect, but I think we need to not pipe to a file/tail it when `HEADLESS=false && IN_CONTAINER`. Furthermore, I believe that we need to do the PID logic regardless of `IN_CONTAINER=true` but when `HEADLESS=true`.

shell/debug.sh
```diff
@@ -83,11 +83,29 @@ if [[ $HEADLESS == "true" ]]; then
   )
 fi
 
+function ctrl_c_trap() {
+  echo "killing delve with PID: $DLV_PID"
+  kill "$DLV_PID"
+  exit
```
tonymbugua-outreach, line 89: nit: I would explicitly exit with code 0

jaredallard: I think we'd actually want `exit 0`, since [by default](https://bash.cyberciti.biz/guide/Exit_command) it'll be the exit code of the `kill` right now.

shell/debug.sh
```diff
@@ -83,11 +85,37 @@ if [[ $HEADLESS == "true" ]]; then
...
+    # a container so the output can be processed by devspace.
+    echo -e "\n\n\n\n\n\n\n\n" >>"$DEV_CONTAINER_LOGFILE"
+    "${delve[@]}" >>"$DEV_CONTAINER_LOGFILE" 2>&1 &
+    IS_LOGGING="true"
+  fi
+
+  DLV_PID=$!
+  echo "delve pid is: $DLV_PID"
+
+  if [[ $IS_LOGGING == "true" ]]; then
```
tonymbugua-outreach, line 115: From Jareds comment I think we want to check if `HEADLESS=false && IN_CONTAINER` before tailing?

joeysmithoutreach: True. Fixed that

tonymbugua-outreach (approved):

PR #551, opened 2023-05-31
fix(prisma): script not retrying upload when artifact not found

tonymbugua-outreach:
## Jira ID

DT-3633

### DIFF
--- a/shell/opslevel/upload.sh
+++ b/shell/opslevel/upload.sh
@@ -21,7 +21,26 @@ source "$DIR/../lib/opslevel.sh"
 
 # Download Artifacts JSON once
 ARTIFACTS_JSON_FILE=/tmp/artifacts.json
-download_artifacts_json "${ARTIFACTS_JSON_FILE}"
+# download_artifacts_json "${ARTIFACTS_JSON_FILE}"
+
+# have a loop that runs every 2 seconds and will terminate after X seconds or if file doesnt contain message
+# Set interval (duration) in seconds.
+SECONDS=20
+# Calculate end timeout.
+ENDTIMEOUT=$(($(date +%s) + SECONDS))
+
+while [ "$(date +%s)" -lt $ENDTIMEOUT ]; do
+  sleep 2
+  download_artifacts_json "${ARTIFACTS_JSON_FILE}"
+  EXITCODE=$?
+  echo "EXIT CODE: ${EXITCODE}"
+
+  if [ "${EXITCODE}" -eq "0" ]; then
+    echo "exit code is zero"
+    break
+  fi
+done
+# capture the result into a var
 
 # Upload scan results to OpsLevel
 # shellcheck source=./upload_prismaci.sh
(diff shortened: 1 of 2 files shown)


shell/opslevel/upload.sh
```diff
@@ -21,7 +21,26 @@ source "$DIR/../lib/opslevel.sh"
...
-download_artifacts_json "${ARTIFACTS_JSON_FILE}"
+
+# have a loop that runs every 2 seconds and will terminate after X seconds or if file doesnt contain message
+# Set interval (duration) in seconds.
+SECONDS=20
+# Calculate end timeout.
+ENDTIMEOUT=$(($(date +%s) + SECONDS))
+
+while [ "$(date +%s)" -lt $ENDTIMEOUT ]; do
+  sleep 2
```
jkinkead, line 32: Putting the sleep after line 40 will make this 2 seconds faster.

jkinkead: You edited the wrong `sleep`. :)

You also still want this - but you want to move it to only happen after failure (line 40 in this file).

shell/opslevel/upload_prismaci.sh
```diff
@@ -23,13 +23,11 @@ if [[ -z ${ARTIFACT_URL} ]]; then
 fi
 
 # Set interval (duration) in seconds.
-SECONDS=10
+SECONDS=20
 # Calculate end timeout.
 ENDTIMEOUT=$(($(date +%s) + SECONDS))
 
 while [ "$(date +%s)" -lt $ENDTIMEOUT ]; do
-  sleep 2
```
jkinkead, line 31: You still want a sleep in this loop.

In this loop, you want it on line 47.

jkinkead (changes requested):

jkinkead (changes requested):
Please add back in the sleep!

shell/opslevel/upload.sh
```diff
@@ -21,7 +21,25 @@ source "$DIR/../lib/opslevel.sh"
...
+  download_artifacts_json "${ARTIFACTS_JSON_FILE}"
+
+  # Check if artifact file is empty
+  if [ -s "${ARTIFACTS_JSON_FILE}" ]; then
+    echo "successfully retrived artifacts json"
+    break
+  else
+    echo "failed to retrive artifacts json"
+    sleep 2
+  fi
```
jaredallard, line 41: nit: since you're `break`-ing, can just omit the `else`.

```suggestion
  fi

   echo "failed to retrive artifacts json"
   sleep 2
```

jaredallard: (may need to run `make fmt` after accepting that)

shell/opslevel/upload_prismaci.sh
```diff
@@ -46,6 +44,7 @@ while [ "$(date +%s)" -lt $ENDTIMEOUT ]; do
   if [[ ${RESULT_FILE_MESSAGE} == "Build not found" ]]; then
     continue
   fi
+  sleep 2
```
jkinkead, line 47: This is the wrong place - you want to `sleep` before you re-loop, which happens at the `continue` on line 45. _That_ used to be line 47 before the sleep delete. :)

You could instead simply remove lines 43-46, since an if-continue at the end of a loop is kind of pointless ...

jkinkead (approved):
LGTM!

PR #556, opened 2023-06-15
feat(devspace): ensure that asdf installs are installed before starting

jaredallard:
Changes the devspace start script to ensure that asdf runtimes are
installed before starting anything (e.g., `make dev`, `make e2e` or
dropping the user into a shell). This reduces the percieved compilation
time since this would otherwise be done lazily on the first `make`
invocation. This is especially important/useful for an upcoming
stencil-golang PR that changes the cache behaviour of `asdf`.

Also changed the bundle authentication to not require `ruby` as `bundle`
is not always available at the time.

### DIFF
--- a/shell/devspace_start.sh
+++ b/shell/devspace_start.sh
@@ -33,10 +33,11 @@ if [[ -n $GH_TOKEN ]]; then
   # shellcheck disable=SC1091
   . "$HOME/.asdf/asdf.sh"
 
-  # Bundler requires an expanded $GH_TOKEN so we use the variable directly here
-  # shellcheck disable=SC2016
-  # shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


shell/devspace_start.sh
```diff
@@ -74,13 +80,13 @@ This is how you can work with it:
 
 if [[ -z $DEV_CONTAINER_LOGFILE ]] || [[ $DEVENV_DEV_TERMINAL == "true" ]]; then
   echo -e "$BANNER"
-  bash
+  exec bash
```
jaredallard, line 83: I forgot to mention this, using `exec` is better for ensuring signals are passed through to things appropriately. Just a minor cleanup.

kaldorn (approved):

PR #560, opened 2023-06-16
fix: don't fail if the field is not an array

joeysmithoutreach:

Prevent yq from throwing an error when there is no array in the given field.

https://stackoverflow.com/questions/28213232/docker-error-jq-error-cannot-iterate-over-null

## Jira ID

[XX-XX]

### DIFF
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -34,7 +34,7 @@ get_image_field() {
   local name="$1"
   local field="$2"
   # shellcheck disable=SC1087 # this is a yq/jq filter that shellcheck thinks is bash
-  yq -r ".[\"$name\"]$field[]" "$MANIFEST"
+  yq -r ".[\"$name\"]$field[]?" "$MANIFEST"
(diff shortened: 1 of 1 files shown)


shell/ci/release/docker.sh
```diff
@@ -34,7 +34,7 @@ get_image_field() {
   local name="$1"
   local field="$2"
   # shellcheck disable=SC1087 # this is a yq/jq filter that shellcheck thinks is bash
-  yq -r ".[\"$name\"]$field[]" "$MANIFEST"
+  yq -r ".[\"$name\"]$field[]?" "$MANIFEST"
```
jaredallard, line 37: TIL, I would've just checked the field type. This is much cooler. You might want to test it with the instructions in the README on a repo w/ and w/o the settings to be safe.

jaredallard:
Superseded by a PR I'm about to open.

jaredallard:
https://github.com/getoutreach/devbase/pull/561

PR #561, opened 2023-06-16
fix: better array parsing, small testing framework, asdf list fix

jaredallard:
Fixes the `docker.sh` array parsing to properly handle an array not
being set. Extracts aforementioned logic to be in a generic `yaml.sh`
library in hopes it can be more usable in the future, since moving to Go
isn't happening yet.

As part of that, I felt we needed some tests here. So, I wrote a _very_
simple testing framework to call `*_test.sh` scripts, much like Go. If
the text returns a non-zero exit code, it is considered failed. Added
some tests for the new logic.

I will add some documenatation for this in the PR shortly, but wanted to
get it out for comments (since it is simple and described here).

Fixed an issue where newer asdf versions seem to put `*` in front of
versions as well as an empty space, so our original `^` regex no longer
works.


bootstrap.lock
jaredallard, line 1: Stale file.

joeysmithoutreach (approved):

shell/ci/release/docker.sh
```diff
@@ -24,17 +24,21 @@ source "${LIB_DIR}/buildx.sh"
...
+# get_image_field is a helper to return a field from the manifest
+# for a given image. It will return an empty string if the field
+# is not set.
 get_image_field() {
-  local name="$1"
+  local image="$1"
   local field="$2"
-  # shellcheck disable=SC1087 # this is a yq/jq filter that shellcheck thinks is bash
-  yq -r ".[\"$name\"]$field[]" "$MANIFEST"
+  yaml_get_array "$(yaml_construct_object_filter "$image" "$field")" "$MANIFEST"
```
ygcao, line 41: not every field (e.g. buildContext) is an array, this introduced a bug.

PR #563, opened 2023-06-20
chore: Promote rc to stable

tonymbugua-outreach:
Promote rc to stable, created by `dt release promote`

**DO NOT MERGE**


jaredallard (changes requested):
Got a bug fix that needs to get into this. Let's hold off.

Also we're not 100% done with release notes yet, so we should put a pause on the releasing until that is done (will update Slack thread)

jaredallard (approved):
bug fix is in :shipit:

PR #568, opened 2023-06-20
fix: don't overwrite prev line if error

joeysmithoutreach:

There is some logic in shell.sh's run_command function to put the terminal back one line. However, this is currently overwriting parts of error messages.

This changes the code so that the cursor is only moved back if the previous command ran successfully.

## Jira ID

[DT-3800]

## Notes for your reviewers

[DT-3800]: https://outreach-io.atlassian.net/browse/DT-3800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/shell.sh
+++ b/shell/lib/shell.sh
@@ -138,7 +138,7 @@ run_command() {
     # If the position of the cursor didn't change, and wasn't zero,
     # we can safely assume that the linter didn't output anything,
     # so we can just overwrite the previous line.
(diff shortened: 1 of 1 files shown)


jaredallard (approved):
Seems like a fair assertion.

PR #571, opened 2023-06-26
chore: Add QF to CODEOWNERS.

jkinkead:


kaldorn (approved):
Adds QF to CODEOWNERS

PR #572, opened 2023-06-26
fix(circleci): use correct version of devbase orb

kaldorn:
Fixes the circleci pipeline.

## Jira ID

[XX-XX]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@dev:local
+  shared: getoutreach/shared@2.14.0
   queue: eddiewebb/queue@1.8.4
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@dev:local
+  shared: getoutreach/shared@2.14.0
```
jaredallard, line 5: Should we have a follow up ticket to fix the template https://github.com/getoutreach/stencil-circleci/blob/main/templates/.circleci/config.yml.tpl#L8?

kaldorn: Yes.

jaredallard (approved):
Fixes CI issue, but would like a ticket to fix the underlying template in the meantime.

PR #576, opened 2023-06-27
fix: devconfig stdout race condition

george-e-shaw-iv:

* fixes a race condition that puts my terminal into raw mode basically
  because two different unsynchronoized processes are writing to stdout
  at the same time? I don't know. It's my best guess. All I know is that
  removing this from running in the background OR not outputting to
  stdout fixes the problem.

## Jira ID

[XX-XX]

## Notes for your reviewers

Example of what happens (This is consistent 100% of the time) when this runs in the background the way it has been the past month or two: https://cl.outreach.io/RBuJ9PRo

### DIFF
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -1,6 +1,9 @@
 // Copyright 2022 Outreach Corporation. All Rights Reserved.
 
-// Description: This file has the package main.
+// Description: This is the entrypoint of the e2e runner for the devenv.
+
+// TODO(george-e-shaw-iv): Remove all calls to log.Fatal with graceful exits in mind.
+
 package main
 
 import (
--- a/e2e/e2e.go
+++ b/e2e/e2e.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"go/build"
 	"os"
 	"os/exec"
@@ -149,10 +150,12 @@ func provisionNew(ctx context.Context, target string) error { // nolint:unparam 
 }
 
 // runDevconfig executes devconfig command
-func runDevconfig(ctx context.Context) {
-	if err := osStdOutErr(exec.CommandContext(ctx, ".bootstrap/shell/devconfig.sh")).Run(); err != nil {
-		log.Fatal().Err(err).Msg("Failed to run devconfig")
+func runDevconfig(ctx context.Context) error {
+	out, err := exec.CommandContext(ctx, "./scripts/shell-wrapper.sh", "devconfig.sh").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%s", out)
 	}
+	return nil
 }
 
 // shouldRunE2ETests denotes whether or not this needs to actually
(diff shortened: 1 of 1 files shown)


e2e/e2e.go
```diff
@@ -314,7 +317,14 @@ func main() { //nolint:funlen,gocyclo // Why: there are no reusable parts to ext
 		go func(wg *sync.WaitGroup) {
 			defer wg.Done()
 			log.Info().Msg("Running devconfig in background")
-			runDevconfig(ctx)
+			if err := runDevconfig(ctx); err != nil {
+				// Call cancel to hopefully communicate a signal back to other currently running commands to stop
+				// doing what they're doing. If we just exit (implicitly via log.Fatal) they likely will continue
```
jkinkead, line 322: It looks like this is using a third-party logger, but it looks like `Fatal` behavior [is the same as the stdlib](https://github.com/rs/zerolog/blob/9070d49a1a4154c819e9c5ae6c923f3cc916d959/log.go#L368).

Since this is calling `os.Exit`, a `cancel` call likely won't have enough time to finish working. Is there a reason you can't just return from this function?

e2e/e2e.go
```diff
@@ -329,7 +339,9 @@ func main() { //nolint:funlen,gocyclo // Why: there are no reusable parts to ext
 
 	if requireDevconfigAfterDeploy {
 		log.Info().Msg("Running devconfig")
-		runDevconfig(ctx)
+		if err := runDevconfig(ctx); err != nil {
+			log.Fatal().Err(err).Msg("failed to run devconfig")
```
jkinkead, line 343: I don't like that this function is using a bunch of `Fatal` calls, and also uses several `defer`s. These will be skipped if this is exiting the process directly.

`Fatal` shouldn't be used in any robust function.

In particular, it _really_ seems like this should be a `return` - but that's also true of all the other `Fatal` lines in this function.

e2e/e2e.go
george-e-shaw-iv: The entirety of this "background runner" is written extremely poorly is the thing. I didn't have a ton of time to sink into fixing this. You're probably right though. I think this needs a little more TLC to work correctly.

e2e/e2e.go
george-e-shaw-iv: You're right - I was just following the current conventions of the file. If we want to leave this open for a little bit I can come back to it and refactor the entire file. I won't have time to do it before July 10th realistically though.

e2e/e2e.go
george-e-shaw-iv: There are a few ways in which we could consistently guarantee that a canceled context coming from a spawned (non-main) go-routine ends up killing all spawned processes (from os.Exec). It's going to require some sort of coordination framework though. Go really doesn't make this easy unfortunately

e2e/e2e.go
jkinkead: Can you throw a TODO at the top of this function to replace the `Fatal` logs with something better, noting the `os.Exit` behavior, and call it good?

jkinkead (approved):

PR #577, opened 2023-07-05
fix: use ci docker build script to build docker images

yj-yan:
We already support building docker images defined in `docker.yaml` file, this change reuse the script to allow `devenv deploy` do the same behavior.

## Jira ID

[DT-3624]

## Notes for your reviewers
I ran local tests for some images, and it does not seem to be a breaking change. I could not run full testing for the `ds-mlp-website` repo, because I could not build the `co-service` successfully. I assume my aws access is not permitted to pull the specific S3 bucket.

[DT-3624]: https://outreach-io.atlassian.net/browse/DT-3624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -202,7 +202,7 @@ docker-build:: pre-docker-build
 	@echo " ===> building docker image <==="
 	@ssh-add -L
 	@echo " ===> If you run into credential issues, ensure that your key is in your SSH agent (ssh-add <ssh-key-path>) <==="
-	@./shell/ci/release/docker.sh
+	@./scripts/shell-wrapper.sh ci/release/docker.sh
 
 ## fmt:             run source code formatters
 .PHONY: fmt
--- a/root/Makefile
+++ b/root/Makefile
@@ -202,7 +202,7 @@ docker-build:: pre-docker-build
 	@echo " ===> building docker image <==="
 	@ssh-add -L
 	@echo " ===> If you run into credential issues, ensure that your key is in your SSH agent (ssh-add <ssh-key-path>) <==="
-	DOCKER_BUILDKIT=1 docker build --ssh default -t gcr.io/outreach-docker/$(APP) -f deployments/$(APP)/Dockerfile . --build-arg VERSION=${APP_VERSION}
+	@./shell/ci/release/docker.sh
 
 ## fmt:             run source code formatters
 .PHONY: fmt
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -142,6 +142,11 @@ build_and_push_image() {
     fi
   fi
 
+  # Build image with registry prefix for devenv
+  if [[ -n "$DEVENV" ]]; then
+      echo "Building Docker Image for devenv"
+      image="gcr.io/outreach-docker/$image"
+  fi
   # Build a quick native image and load it into docker cache for security scanning
   # Scan reports for release images are also uploaded to OpsLevel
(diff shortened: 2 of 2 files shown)


joeysmithoutreach (approved):

root/Makefile
```diff
@@ -202,7 +202,7 @@ docker-build:: pre-docker-build
 	@echo " ===> building docker image <==="
 	@ssh-add -L
 	@echo " ===> If you run into credential issues, ensure that your key is in your SSH agent (ssh-add <ssh-key-path>) <==="
-	DOCKER_BUILDKIT=1 docker build --ssh default -t gcr.io/outreach-docker/$(APP) -f deployments/$(APP)/Dockerfile . --build-arg VERSION=${APP_VERSION}
+	@./shell/ci/release/docker.sh
```
malept, line 205: issue: I actually tried this change out locally because I was having trouble building a docker image in my service, and I got the following error:

```
bash: line 1: ./shell/ci/release/docker.sh: No such file or directory
```

This is what worked for me to at least get the script running:
```suggestion
		@./scripts/shell-wrapper.sh ci/release/docker.sh
```

shell/ci/release/docker.sh
```diff
@@ -142,10 +142,16 @@ build_and_push_image() {
...
+  # Build image with registry prefix for devenv
+  if [[ -n $DEVENV_DOCKER_BUILD ]]; then
+    info "Building Docker Image for devenv"
+    image="gcr.io/outreach-docker/$image"
   # Build a quick native image and load it into docker cache for security scanning
   # Scan reports for release images are also uploaded to OpsLevel
   # (test image reports only available on PR runs as artifacts).
-  info "Building Docker Image (for scanning)"
+  else
+    info "Building Docker Image (for scanning)"
```
malept, line 153: Nit: I would move the comments into the `else` block, right before the `info` log and at the same indentation as `info` (GitHub won't let me make a suggestion for that).

root/Makefile
yj-yan: Thanks for testing this out. Wired, I tried this locally and works for me. Might be different indentation messed up? I use zsh. I will check the indentation again.

shell/ci/release/docker.sh
yj-yan: ++, I just pushed an update for fixing this. You could be editing in the gap.

root/Makefile
yj-yan: I updated to use shell-wrapper to run the script, indentation doesn't seem to matter for me. LMK if it still doesn't work for your environment.

malept: Whoops, the indentation was a copy/paste error. Sorry about that.

shell/ci/release/docker.sh
malept: Sorry, this is what I meant (manual suggestion block)

```suggestion
  else
    # Build a quick native image and load it into docker cache for security scanning
    # Scan reports for release images are also uploaded to OpsLevel
    # (test image reports only available on PR runs as artifacts).
    info "Building Docker Image (for scanning)"
```

yj-yan: ahh, good catch, thanks. Github didn't let me commit from the here directly, I manually modified it as suggestion.

PR #579, opened 2023-07-05
fix: Performance improvements for asdf.sh.

jkinkead:
Only reshim if we installed new tools. This saves 10+ seconds when you have a lot of tool versions installed.

Only run `awk` once, instead of two times per tool. This saves another 100ms or so.

---

`make` startup is ridiculously slow. This helps a lot with that.

### DIFF
--- a/shell/lib/asdf.sh
+++ b/shell/lib/asdf.sh
@@ -15,11 +15,11 @@ asdf_plugins_list_regenerate
 
 # read_all_asdf_tool_versions combines all .tool-versions found in this directory
 # and the child directories minus node_modules and vendor.
-# Then strip the comments and run uniq.
+# This prints the plugin, then the version, each separated by a newline.
 read_all_asdf_tool_versions() {
   find . -type d \( -path ./.git -o -path ./vendor -o -path ./node_modules \) -prune -o \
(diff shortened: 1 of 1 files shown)


shell/lib/asdf.sh
```diff
@@ -69,13 +69,14 @@ asdf_devbase_exec() {
 # .tool-version files in the current directory and all subdirectories.
 asdf_devbase_ensure() {
   readarray -t asdf_entries < <(read_all_asdf_tool_versions)
-  for entry in "${asdf_entries[@]}"; do
+  local need_reshim=0
+  for (( i=0; i < "${#asdf_entries[@]}"; i=(($i + 2)) )); do
```
malept, line 73: thought: 😬 if only there was a better way to read pairs of words...

malept (approved):
Looks like reshims only happen when a new version of a tool happens.

shell/lib/asdf.sh
jkinkead: I mean, if I wasn't lazy I would've turned this into a map instead of an array ("the bashies struct"), but that would've required rewriting more code. :P

PR #584, opened 2023-07-12
fix: GOOS handling

marekfexa-outreach:
- When introducing devspace binary sync new bug was introduced that causes gobin/air to be build for linux on macos dev machine
- This PR is about fixing this and setting git tag version of air

## Jira ID

[QF-911]


root/Makefile
```diff
@@ -34,6 +34,7 @@ LOG                 := ./scripts/shell-wrapper.sh makefile-logger.sh
 FLY                 := $(CURDIR)/scripts/shell-wrapper.sh fly.sh
 BOX_CONFIG          := ./scripts/shell-wrapper.sh boxconfig.sh
 CONCOURSE_URL       := $(shell $(BOX_CONFIG) '.cd.concourse.address')
+AIR   							:= $(CURDIR)/scripts/shell-wrapper.sh gobin.sh github.com/cosmtrek/air@v1.44.0 -c $(CURDIR)/.air.toml
```
tomassusovsky-outreach, line 37: spacing seems a bit off here?

tomassusovsky-outreach (approved):

PR #589, opened 2023-07-14
feat: Allow timeout configuration in shared/e2e

Erik-Outreach:
For when there are e2e tests which are expected to take longer than 1 min.

[RMS-2967]

### DIFF
--- a/orbs/shared/jobs/e2e.yaml
+++ b/orbs/shared/jobs/e2e.yaml
@@ -16,11 +16,16 @@ parameters:
     description: The timeout that gets applied when CircleCI receives no output during the running of e2e tests.
     type: string
     default: 10m
+  go_test_timeout:
+    description: Maps to gotest -timeout parameter.
+    type: string
+    default: ""
 executor:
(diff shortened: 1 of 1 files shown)


jaredallard (approved):

Erik-Outreach:
@jaredallard thanks for the approval! Now how do we deploy a new version of devbase? 🤔

Erik-Outreach:
> You ask us! :D
> 
> This wouldn't go out until next month though, or we can promote it to `rc` and get it out with the stable release on Monday. If it's critical.

It's not critical but nice to have as it will solve annoyances in RMS tests (such as having 100+ failed tests when it should only display a few) it's a minor impact change so I'd propose to promote to rc

Erik-Outreach:
> You ask us! :D
> 
> This wouldn't go out until next month though, or we can promote it to `rc` and get it out with the stable release on Monday. If it's critical.

@jaredallard if we can promote it to rc, it will help to solve this problem on Monday: https://github.com/getoutreach/rms/pull/1349

PR #597, opened 2023-07-20
fix: Don't echo comment.

jkinkead:
maek


jaredallard (approved):
it was fun while it lasted

PR #598, opened 2023-07-20
feat: rewrite test.sh to be better documented and use bash arrays

jaredallard:
Changed `test.sh` to fully document all environment variables in use.
Changed `_FLAG` and `_TAG` environment variables to use bash arrays to
ensure consistent space handling behaviour, while keeping space
splitting for pre-specified environment variables for compatability.

When passing `-v` to `TEST_FLAGS` (and in turn to `go test`) we now set
the `gotestsum` output format to be `standard-verbose` which is
equivalent to `go test -v`[1] output. Before, this was silently ignored
and frustrating to work with.

Updated the `shared/setup_docker` step to use a newer version of Docker
as the older version had issues with apt package installs (somehow?)

[1]: https://github.com/gotestyourself/gotestsum#:~:text=standard%2Dverbose%20%2D%20the%20standard%20go%20test%20%2Dv%20format.

### DIFF
--- a/shell/test.sh
+++ b/shell/test.sh
@@ -144,7 +144,7 @@ if [[ "$(git ls-files '*_test.go' | wc -l | tr -d ' ')" -gt 0 ]]; then
     # If we have -v passed as a test flag, we use the standard-verbose
     # format which is equal to 'go test -v'. Without this, we would
     # silently ignore the flag.
-    if [[ "$flag" == "-v" ]]; then
+    if [[ $flag == "-v" ]]; then
       format="standard-verbose"
     fi
   done
@@ -154,7 +154,10 @@ if [[ "$(git ls-files '*_test.go' | wc -l | tr -d ' ')" -gt 0 ]]; then
 
   # Convert TEST_TAGS into the expected format for `go test -tags`,
   # which is a comma-separated string.
-  test_tags_string=$(IFS=","; echo "${TEST_TAGS[*]}")
+  test_tags_string=$(
+    IFS=","
+    echo "${TEST_TAGS[*]}"
+  )
 
   if [[ -n $PACKAGE_TO_DEBUG ]]; then
     TESTBIN=$(mktemp)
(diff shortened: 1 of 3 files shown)


shell/test.sh
```diff
@@ -1,114 +1,193 @@
 #!/usr/bin/env bash
+# Runs tests for the current project.
+set -euo pipefail
```
jaredallard, line 3: Enabled `e` (exit on error) `u` (error on undefined env vars) `o pipefail` (fail on pipe errors). It makes bash scripts much more resilient, and also a forcing function to document env vars :D

shell/test.sh
```diff
@@ -1,114 +1,193 @@
...
-  COVER_FLAGS=${COVER_FLAGS:- -coverpkg=./... -covermode=atomic -coverprofile=/tmp/coverage.out -cover}
+# If WITH_COVERAGE is set, we pass flags to 'go test' to enable coverage
+# file generation.
+if [[ -n $WITH_COVERAGE ]]; then
+  COVER_FLAGS+=(-coverpkg=./... -covermode=atomic -coverprofile=/tmp/coverage.out -cover)
 fi
 
 testInclude="$(get_repo_directory)/scripts/test.include.sh"
 if [[ -e $testInclude ]]; then
+  info "Running project specific test file: $testInclude"
```
jaredallard, line 125: This should help make it more clear to a user when a project-specific test file is failing vs the devbase framework.

versions.yaml
```diff
@@ -2,7 +2,7 @@ grpcui: 1.3.1
 jsonnetfmt: 0.19.1
 goimports: 0.5.0
 delve: 1.20.1
-gotestsum: 1.9.0
+gotestsum: 1.10.1
```
jaredallard, line 5: Updated `gotestsum` because it didn't seem to break anything.

kaldorn (approved):
This documents the flags better.

PR #599, opened 2023-07-20
fix: only publish node client if releasing a service

joeysmithoutreach:

This change prevents a publishing node clients if the repo is not a service and therefore will not have any clients to publish.

## Jira ID

[DT-3704]

## Notes for your reviewers

[DT-3704]: https://outreach-io.atlassian.net/browse/DT-3704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jaredallard (approved):
Seems reasonable. Alternatively could check if that path exists since it's less assuming of stencil-golang generating things, but we're going to nuke this code not that long from now (post-migration to github packages for npm) so I'm OK with this.

PR #600, opened 2023-07-20
feat: allow per project linters

evadnoob:
These changes allow per-project linters to be defined in a well known location. If the location exists(dir) and linters are found, they are added to the linters list.


shell/linters.sh
```diff
@@ -17,23 +17,32 @@ if [[ -n $SKIP_LINTERS ]] || [[ -n $SKIP_VALIDATE ]]; then
   exit 0
 fi
 
+# add extra, per project, linters
+linters=("$DIR/linters"/*.sh)
+if [[ -z $workspacefolder ]]; then
+    workspacefolder="$(get_repo_directory)"
+fi
```
malept, line 24: question: what is the case where `$workspacefolder` is not empty by this time in the script?

evadnoob: This was just following existing style/conventions used in other scripts, [here](https://github.com/getoutreach/devbase/blob/1c78e07d657de9cb9c1623b76a9e26745bf4e444/shell/golangci-lint.sh#L14) and [here](https://github.com/getoutreach/devbase/blob/1c78e07d657de9cb9c1623b76a9e26745bf4e444/shell/golangci-lint.sh#L13)

I kept that convention, didn't want to break pre-existing expectation

masterkidan (approved):

shell/linters.sh
```diff
@@ -17,23 +17,30 @@ if [[ -n $SKIP_LINTERS ]] || [[ -n $SKIP_VALIDATE ]]; then
   exit 0
 fi
 
+# add extra, per project, linters
```
malept, line 20: Nit: I think the commas are a bit excessive
```suggestion
# add extra (per project) linters
```

shell/linters.sh
```diff
@@ -17,23 +17,30 @@ if [[ -n $SKIP_LINTERS ]] || [[ -n $SKIP_VALIDATE ]]; then
...
+linters=("$DIR/linters"/*.sh)
+if [[ -z $workspaceFolder ]]; then
+  workspaceFolder="$(get_repo_directory)"
+fi
+if [[ -d "$workspaceFolder"/scripts/linters ]]; then
+  linters+=("${workspaceFolder}"/scripts/linters/*.sh)
```
malept, line 26: Nit: I don't think the curly braces are necessary.
```suggestion
  linters+=("$workspaceFolder"/scripts/linters/*.sh)
```

malept (commented):
This seems fine but I'd like @jaredallard to review the high level approach.

shell/linters.sh
jaredallard: It's personal styl pref of mine to always quote the entire string unless needed. So, I would also say this should be 

```suggestion
  linters+=("$workspaceFolder"/scripts/linters/"*.sh)
```

jaredallard (commented):
This LGTM, but it'd be nice to have a docs entry in [docs/linters.md](https://github.com/getoutreach/devbase/blob/main/docs/linters.md) describing this.

evadnoob:
> This LGTM, but it'd be nice to have a docs entry in [docs/linters.md](https://github.com/getoutreach/devbase/blob/main/docs/linters.md) describing this.

Added, thanks for the suggestion.

jaredallard:
> > This LGTM, but it'd be nice to have a docs entry in [docs/linters.md](https://github.com/getoutreach/devbase/blob/main/docs/linters.md) describing this.
> 
> Added, thanks for the suggestion.

I think you forgot to push?

docs/linters.md
```diff
@@ -10,3 +10,14 @@ There is currently no way to ignore files outside of the `.gitignore` file.
 
 There is currently a bash issue where the exit code is not getting detected when there is multiple commands being ran.
 The easy solution for now is to return the value 1 from the function indicating an error has occurred and terminating the `make test/lint` command that triggered it
+
+## Migrations Linter ##
```
jaredallard, line 14: Ah, I more meant adding documentation on the linter framework that was added. The migrations linter also makes sense (for Outreach internal docs), but here I think we should briefly document the per-project linters!

PR #605, opened 2023-07-27
feat: Remove OpsLevel and Prismacloud integration

jaredallard:
Removed the OpsLevel deployment notifications and maturity checking that
blocked CLI releases. Removed prismacloud security scanning and
subsequent results uploading to OpsLevel logic. Adjusted the `docker.sh`
builder to make it more clear why we still have different `docker build`
invocations.

These integrations weren't really what we wanted for devbase, but also
we're no longer using/supporting these at Outreach. So, we remove them
in hopes that we'll add something else back that's better (or not at
all!).


.circleci/config.yml
```diff
@@ -14,7 +14,6 @@ parameters:
 contexts: &contexts
   - aws-credentials
   - prismacloud-credentials
-  - opslevel-credentials
```
jaredallard, line 17: Stencil did it.

shell/ci/release/docker.sh
```diff
@@ -164,6 +148,14 @@ build_and_push_image() {
         -t "$remote_image_name:$VERSION" -t "$remote_image_name:latest" --push \
         "$buildContext"
     )
+  else
+    # When we're not on a tag, just build the docker image to verify
+    # that it is buildable.
+    info "Building Test Docker Image"
+    (
+      set -x
+      docker buildx build "${args[@]}" -t "$image" --load "$buildContext"
```
jkinkead, line 157: Any reason you're omitting `--platform`?

Maybe store common args in an array ... ?

shell/ci/release/docker.sh
```diff
@@ -164,6 +148,14 @@ build_and_push_image() {
         -t "$remote_image_name:$VERSION" -t "$remote_image_name:latest" --push \
         "$buildContext"
     )
+  else
+    # When we're not on a tag, just build the docker image to verify
+    # that it is buildable.
+    info "Building Test Docker Image"
```
jkinkead, line 154: I don't love this, because it implies that the image is used for testing. "Running Test Docker Build" maybe? "Verifying Docker Build"?

jkinkead (approved):
LGTM, a couple small nits.

jkinkead:
y u automerge

shell/ci/release/docker.sh
jaredallard: I should've expanded in a comment here. We omit `platform` because we want to do a quick validation. Though, I guess building for both platforms is a good idea, in retrospect.

shell/ci/release/docker.sh
jaredallard: I'll address this in a follow-up. Hard agree. I didn't like the phrasing much either.

PR #606, opened 2023-07-27
fix: update docker.sh statement, deduplicate logic

jaredallard:
As a follow up to #605, this updates the docker.sh output to better
reflect what the docker build is doing.

I also took the time to combine the logic for the docker build _test_
logic and the docker build _release_ logic (determined by the presence
of `CIRCLE_TAG`. This brings the added benefit of us building a docker
image on all archs that we would release to on PRs.

Updated buildx to reflect what is in use on most user's machines.

### DIFF
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -63,6 +63,9 @@ get_image_field() {
 build_and_push_image() {
   local image="$1"
 
+  # push determines if we should push the image to the registry or not.
+  local push=false
+
   # Platforms to build this image for, expected format (in YAML):
   # platforms:
   #   - linux/arm64
@@ -138,6 +141,11 @@ build_and_push_image() {
   tags=()
   if [[ -n $CIRCLE_TAG ]]; then
     tags+=("$remote_image_name:$CIRCLE_TAG" "$remote_image_name:latest")
+
+    # When on a tag, we should also push the image to the registry. Also
+    # set push to true so that we log the right message.
+    args+=("--push")
+    push=true
   else
     tags+=("$image")
   fi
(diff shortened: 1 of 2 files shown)


shell/ci/release/docker.sh
```diff
@@ -139,24 +159,26 @@ build_and_push_image() {
...
-    )
+  if [[ $push == true ]]; then
+    echo "🔨 Building and Pushing Docker Image"
   else
-    # When we're not on a tag, just build the docker image to verify
-    # that it is buildable.
-    info "Building Test Docker Image"
-    (
-      set -x
-      docker buildx build "${args[@]}" -t "$image" --load "$buildContext"
```
jaredallard, line 157: Note: now that we're not scanning the image, we no longer need `--load`. So, I removed it. This should be marginally faster. But also, we can't load multi-arch docker images currently.

shell/ci/release/docker.sh
```diff
@@ -129,6 +129,26 @@ build_and_push_image() {
...
+  if [[ -n $CIRCLE_TAG ]]; then
+    tags+=("$remote_image_name:$CIRCLE_TAG" "$remote_image_name:latest")
+  else
+    tags+=("$image")
+  fi
+  for tag in "${tags[@]}"; do
+    args+=("--tag" "$tag")
+  done
+
+  # When on a tag, we should push the image to the registry.
```
jkinkead, line 148: Put this & body in `if` on line 139?

shell/ci/release/docker.sh
```diff
@@ -139,24 +159,27 @@ build_and_push_image() {
       buildContext="$(get_repo_directory)/deployments/$image"
     fi
   fi
+  args+=("$buildContext")
+
+  # Check if the arguments include --push, if so, we'll push
```
jkinkead, line 164: Can you set when you append the argument, instead of doing a loop a few lines below?

jkinkead (approved):
Mostly LGTM.

shell/ci/release/docker.sh
jaredallard: I didn't like mixing the tags logic... but yeah ig it makes sense.

shell/ci/release/docker.sh
jaredallard: I wanted to detach from the CIRCLE_TAG check in case we moved away from it, but yeah, that's reasonable.

jaredallard:
Tested the PR validation part, seems to work great: https://app.circleci.com/pipelines/github/getoutreach/discovery/685/workflows/49a79f7a-fe7f-418d-b221-22e501a351c1/jobs/2321

malept (approved):

PR #616, opened 2023-08-03
fix: add missing per-project formatter for migrations

evadnoob:
adds per-project formatters, just like linters

### DIFF
--- a/shell/fmt.sh
+++ b/shell/fmt.sh
@@ -13,7 +13,6 @@ source "$DIR/lib/bootstrap.sh"
 # shellcheck source=./lib/shell.sh
 source "$DIR/lib/shell.sh"
 
-
 # add extra (per project) linters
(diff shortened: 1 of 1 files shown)


malept (commented):
Do we still want to call have `linter` in the variable names despite it being for formatters?

evadnoob:
> Do we still want to call have `linter` in the variable names despite it being for formatters?

I kept it the same as the other script(the linters.sh) and there was already a reference to the "linters" directory. Also the linters file is what has the formatter function. I'm on the fence about this but leaning toward leaving it as linters, since that's where these formatters come from.

malept (approved):
I don't particularly like the current naming scheme, since linters and formatters are effectively a venn diagram, but it's not something I'm interested in investing more energy arguing about.

PR #619, opened 2023-08-07
fix(confluence): don't append logs to markdown output

malept:

Avoids redirecting `stdout` of `visualizemd` to create the markdown file, and instead uses the existing `-o` flag.

## Jira ID

[DT-3756]

[DT-3756]: https://outreach-io.atlassian.net/browse/DT-3756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jaredallard (approved):
Even if this doesn't fix anything, seems reasonable.

PR #622, opened 2023-08-08
fix(linters): skip kubecfg linter if app jsonnet doesn't exist

malept:

This fixes the case where the kubecfg linter is run in a Stencil repo that isn't a service but has jsonnet files for whatever reason.


jaredallard (approved):
I was wondering if we'd want to do this or not. Seems reasonable.

PR #623, opened 2023-08-08
feat(golangci-lint): Upgrade to 1.53.3

jaredallard:
Upgrades `golangci-lint` to `1.53.3` to support Go 1.20.

This likely brings in some new lint rules that _may_ fail on new
projects. Putting it in unstable to give us sometime to run into them
and adjust our rules as needed.

I didn't hit any in a few repos I tried this on.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -4,7 +4,7 @@
 # you are reducing compatibility guarantees.
 ## <<Stencil::Block(toolverOverride)>>
 ## <</Stencil::Block>>
-golang 1.19.11
+golang 1.20.7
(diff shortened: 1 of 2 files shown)


malept:
Merge before or after we cut the RC?

jaredallard:
@malept After. I'll wait until that's fixed!

kaldorn:
I think you'll also want to disable depguard. 
Ref: https://github.com/getoutreach/devbase/pull/552

kaldorn (approved):

PR #625, opened 2023-08-11
feat(auth/github_packages): support setting GITHUB_TOKEN

jaredallard:
Currently, you can manually specify `GITHUB_TOKEN` for `auth/github` but
not for `auth/github_packages`. While, I would like to make calling
`auth/github` just a requirement for `auth/github_packages`, that seems
breaking. So, instead, we just match the behaviour of `auth/github` and
allow using `GITHUB_TOKEN` if set.

This is useful for repos that manually use pieces of devbase (for
example, Outreach's flagship) that may not have things like Go installed
yet.

### DIFF
--- a/shell/ci/auth/github_packages.sh
+++ b/shell/ci/auth/github_packages.sh
@@ -7,9 +7,12 @@ LIB_DIR="${DIR}/../../lib"
 # shellcheck source=../../lib/bootstrap.sh
 source "${LIB_DIR}/bootstrap.sh"
 
-GH_TOKEN=$("$SHELL_DIR/gobin.sh" \
-  "github.com/getoutreach/ci/cmd/ghaccesstoken@$(get_tool_version "getoutreach/ci")" \
-  --skip-update token --env-prefix "GHACCESSTOKEN_PAT")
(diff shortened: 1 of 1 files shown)


shell/ci/auth/github_packages.sh
```diff
@@ -7,9 +7,12 @@ LIB_DIR="${DIR}/../../lib"
 # shellcheck source=../../lib/bootstrap.sh
 source "${LIB_DIR}/bootstrap.sh"
 
-GH_TOKEN=$("$SHELL_DIR/gobin.sh" \
-  "github.com/getoutreach/ci/cmd/ghaccesstoken@$(get_tool_version "getoutreach/ci")" \
-  --skip-update token --env-prefix "GHACCESSTOKEN_PAT")
+# Fetch the token from ghaccesstoken if not set.
+if [[ -z $GITHUB_TOKEN ]]; then
```
jaredallard, line 11: Renamed `GH_TOKEN` -> `GITHUB_TOKEN` to match `auth/github` as well.

malept (approved):

PR #626, opened 2023-08-11
fix: Upgrading goreleaser to 1.20.0 to fix releases

deregtd:
Older versions of goreleaser have a bug that detects go v1.20 incorrectly and stops releasing darwin/arm64.  We have to upgrade to fix it. https://github.com/goreleaser/goreleaser/issues/3835


jaredallard:
Well, that's hilarious. Good catch!

PR #630, opened 2023-08-14
fix(golangci-lint): upgrade to v1.54.1, fail on invalid version pairs

jaredallard:
Upgrades golangci-lint to v1.54.1 to support Go 1.21. Adds a small
framework in `golangci-lint.sh` to explode if versions of Go and
golangci-lint are not compatible. Adds checks just for 1.21 and 1.20.
This is important because, most notably for 1.20, it can explode in CI
by consuming all the RAM and not be clear why that is happening.

Changes `GOGC` to no longer be set in favor of using total memory
calculation to pass to `GOMEMLIMIT`. Fallsback to `GOGC` if required.
It's important to note that this doesn't disable GC, just makes it less
important until we get close to total memory (or are otherwise unable to
free any).

Also updated `linter.sh`'s error to not blindly suggest `make fmt`.
Instead, it recommends it based on the output.

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -92,7 +92,8 @@ release:: pre-release
 	@# Create a tag for our version
 	@git tag -d "$(APP_VERSION)" >&2 >/dev/null || true
 	@git tag "$(APP_VERSION)" >&2
-	@GORELEASER_CURRENT_TAG=$(APP_VERSION) goreleaser release --skip-announce --skip-publish --skip-validate --rm-dist
+	@GORELEASER_CURRENT_TAG=$(APP_VERSION) ASDF_GORELEASER_VERSION=$(shell cat "$(BOOTSTRAP_DIR)/.tool-versions" | grep goreleaser | awk '{ print $$2 }') \
+		goreleaser release --skip-announce --skip-publish --skip-validate --rm-dist
 	@# Delete the tag once we are done.
 	@git tag -d "$(APP_VERSION)" >&2
 
--- a/shell/golangci-lint.sh
+++ b/shell/golangci-lint.sh
@@ -23,37 +23,53 @@ GOLANGCILINT_VERSION=$(asdf_devbase_run golangci-lint --version | awk '{print $4
 GO_MINOR_VERSION_INT=${GO_MINOR_VERSION//./}
 GOLANGCI_LINT_VERSION_INT=${GOLANGCILINT_VERSION//./}
 
-# Go 1.20 requires >= golangci-lint 1.52.0
-if [[ $GO_MINOR_VERSION_INT == 120 ]] && [[ $GOLANGCI_LINT_VERSION_INT -lt 1520 ]]; then
-  echo "Error: Go 1.20 requires golangci-lint 1.52.0 or newer (detected $GOLANGCILINT_VERSION)" >&2
-  exit 1
-fi
-
-# Go 1.21 requires >= golangci-lint 1.54.1
-if [[ $GO_MINOR_VERSION_INT == 121 ]] && [[ $GOLANGCI_LINT_VERSION_INT -lt 1541 ]]; then
-  echo "Error: Go 1.21 requires golangci-lint 1.54.1 or newer (detected $GOLANGCILINT_VERSION)" >&2
-  exit 1
-fi
-
-# If we're on a system with free, set GOMEMLIMIT to a value that's less
-# than the max amount of RAM on the system. This helps ensure that we
-# don't go over the memory limit and get OOM killed. This is mostly
-# important for CI systems.
-if (command -v free || command -v sysctl) &>/dev/null; then
-  if command -v free &>/dev/null; then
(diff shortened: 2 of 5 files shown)


shell/golangci-lint.sh
```diff
@@ -17,8 +17,45 @@ fi
...
 
-# trade memory for CPU when running the linters
-export GOGC=20
+# Determine the version of go and golangci-lint to calculate compatibility.
+GO_MINOR_VERSION=$(go version | awk '{print $3}' | sed 's/go//' | cut -d'.' -f1,2)
+GOLANGCILINT_VERSION=$(asdf_devbase_run golangci-lint --version | awk '{print $4}')
+GO_MINOR_VERSION_INT=${GO_MINOR_VERSION//./}
+GOLANGCI_LINT_VERSION_INT=${GOLANGCILINT_VERSION//./}
+
+# Go 1.20 requires >= golangci-lint 1.52.0
```
jaredallard, line 26: I guess this type of checking isn't as safe as I'd like it to be... since it'll break on a 2.0 version 😢 

I feel like I could just make a note of that here and remove it when that happens.

malept: Add a `$GO_MINOR_VERSION_INT[0] < 2` check or however it is you just check the major version?

shell/golangci-lint.sh
```diff
@@ -17,8 +17,45 @@ fi
...
+  exit 1
+fi
+
+# Go 1.21 requires >= golangci-lint 1.54.1
+if [[ $GO_MINOR_VERSION_INT == 121 ]] && [[ $GOLANGCI_LINT_VERSION_INT -lt 1541 ]]; then
+  echo "Error: Go 1.21 requires golangci-lint 1.54.1 or newer (detected $GOLANGCILINT_VERSION)" >&2
+  exit 1
+fi
+
+# If we're on a system with free, set GOMEMLIMIT to a value that's less
```
malept, line 38: Nit:
```suggestion
# If we're on a system with free or sysctl, set GOMEMLIMIT to a value that's less
```

shell/golangci-lint.sh
```diff
@@ -17,8 +17,45 @@ fi
...
+# don't go over the memory limit and get OOM killed. This is mostly
+# important for CI systems.
+if (command -v free || command -v sysctl) &>/dev/null; then
+  if command -v free &>/dev/null; then
+    mem="$(free -m | awk '/^Mem:/{print $2}')"
+  elif command -v sysctl &>/dev/null; then
+    mem=$((($(sysctl -n hw.memsize) / 1024) / 1024))
+  fi
+
+  if [[ $mem -lt 1024 ]] || [[ -z $mem ]]; then
```
malept, line 49: suggestion: make it easier to tweak the magic number
```suggestion
  RESERVED_MEMORY_IN_MIB=1024
  if [[ $mem -lt $RESERVED_MEMORY_IN_MIB ]] || [[ -z $mem ]]; then
```

shell/golangci-lint.sh
```diff
@@ -17,8 +17,45 @@ fi
...
+    # Failed to determine GOMEMLIMIT somehow. Fallback to GOGC.
+    echo "Warning: Failed to determine system memory or under threshold. " \
+      "Falling back to GOGC" >&2
+    export GOGC=20
+  else
+    # Use mem as the memory target and ensure that we have 1GB of room.
+    export GOMEMLIMIT="$((mem - 1024))MiB"
```
malept, line 56: suggestion:
```suggestion
    export GOMEMLIMIT="$((mem - $RESERVED_MEMORY_IN_MIB))MiB"
```

shell/golangci-lint.sh
jaredallard: I was thinking about that. Good call :D

root/Makefile
```diff
@@ -92,7 +92,8 @@ release:: pre-release
 	@# Create a tag for our version
 	@git tag -d "$(APP_VERSION)" >&2 >/dev/null || true
 	@git tag "$(APP_VERSION)" >&2
-	@GORELEASER_CURRENT_TAG=$(APP_VERSION) goreleaser release --skip-announce --skip-publish --skip-validate --rm-dist
+	@GORELEASER_CURRENT_TAG=$(APP_VERSION) ASDF_GORELEASER_VERSION=$(shell cat "$(BOOTSTRAP_DIR)/.tool-versions" | grep goreleaser | awk '{ print $$2 }') \
```
jaredallard, line 95: I accidentally snuck this in here, this fixes a small break introduced in #628. Given it's accidentally here... I'm leaning toward just going with it.

malept: I'm assuming this is here because of current working directory shenanigans?

malept (approved):

root/Makefile
jaredallard: Correct!

PR #633, opened 2023-08-15
fix(orbs/save_cache): support setup_remote_docker

jaredallard:
Right now, `test` supports `setup_remote_docker` but `save_cache` does
not. This leads to potential issues with tests that need a remote docker
daemon to be available for their tests (thus failing to run on
`save_cache` but _not_ the `test` job).

This, fixes that, by supporting it in `save_cache` as well.


jaredallard:
I published this to `dev:first` for now so I can unblock https://github.com/getoutreach/orc/pull/891

malept (approved):
Looks legit

PR #636, opened 2023-08-15
test: move from custom framework to bats

jaredallard:
Moves away from our custom bash test from our custom bash testing
framework to instead use `bats`[1]. This is the best currently available
bash testing framework available. This enables us to better test
functionality in bash that we expose today in devbase, until we've moved
more things to Go.

Moved all of the existing tests to use the bats format, as well as
enabled `.bats` linting and formating support through the already
configured shellcheck and shellfmt respectively.

Added documentation to `CONTRIBUTING.md` on how to write these tests.
Added a VSCode plugin recommendation for syntax highlighting.

Fixed the test hook to actually run these tests in CI.

[1]: https://github.com/bats-core/bats-core


scripts/bash-test-runner.sh
```diff
@@ -2,20 +2,20 @@
 # Runs all files with _test.sh at the end of the filename
```
malept, line 2: Nit:
```suggestion
# Runs all files with .bats at the end of the filename
```

malept (approved):
Seems good, just a nit

tomassusovsky-outreach (approved):

PR #637, opened 2023-08-16
feat(machine): support newer versions of Ubuntu

jaredallard:
Adds support for newer versions of Ubuntu and Debian by using the `yq`
apt package, if it's available, and falling back to `pip3` if it's not.

Cleans up the code a bit to make it easier to determine what version of
`gh` is in use/desired. Updates to `v2.32.1`. This will be used by
`bootstrap/dev` in an upcoming PR.

### DIFF
--- a/shell/circleci/machine.sh
+++ b/shell/circleci/machine.sh
@@ -3,6 +3,17 @@
 # These are usually already installed in a CircleCI docker image.
 set -e
 
+# ARCH is the current architecture of the machine. Valid values are:
+#   - amd64
+#   - arm64
+ARCH="amd64"
+if [[ "$(uname -m)" == "aarch64" ]]; then
+  ARCH="arm64"
+fi
+
+# GH_VERSION is the version of gh to install.
+export GH_VERSION=2.32.1
+
 # should_install_vault is a helper function that checks if the vault
 # binary is already installed, if so it returns false. Otherwise, it
 # returns true.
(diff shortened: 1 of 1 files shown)


shell/circleci/machine.sh
```diff
@@ -21,40 +32,17 @@ fi
 # Add APT repositories we do need.
 if should_install_vault; then
   wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o hashicorp-archive-keyring.gpg
-
-  # 2023-02-22: Hashicorp updated their keyring but removed their older key, this reulted in no
```
jaredallard, line 25: Removed this since the parent issue has been closed/resolved.

malept:
Doesn't this repo control which CircleCI executor is used? Or is supporting older Debian/Ubuntu versions because it's configurable in the repo's generated ci.yml?

jaredallard:
> Doesn't this repo control which CircleCI executor is used? Or is supporting older Debian/Ubuntu versions because it's configurable in the repo's generated ci.yml?

It generally does! However, it can be configured. Also, it's used in some Docker images to handle bootstrapping the image (see https://github.com/getoutreach/outreach-docker/pull/682)

shell/circleci/machine.sh
```diff
@@ -21,52 +32,41 @@ fi
...
+
+  echo "Failed to install yq through apt, falling back to pip3 install (this only works on Ubuntu 22.04 and below)"
+  sudo pip3 install yq
+fi
 
-# Vault
 if should_install_vault; then
   echo "Installing Vault"
   sudo apt-get install -y vault
+  sudo rm -rf /opt/vault
```
malept, line 71: Question: what does this delete?

malept (approved):
This seems fine, just one question.

shell/circleci/machine.sh
jaredallard: Ah, this just cleans up the vault server TLS certs and whatnot that were already generated. Seemed better than leaving it around.

PR #639, opened 2023-08-16
fix(release): always ensure PR metadata is omitted

jaredallard:
Right now, due to our release process, we need to open PRs to merge
commits back into the correct branches. Because of this, sometimes a
closed PR will refuse to release.

To prevent this from happening, we remove those environment variables.

This fixes the current issue, I'll have a PR to refactor the existing
logic for this out soon.


kaldorn (approved):

malept (approved):
Looks legit

PR #644, opened 2023-08-18
fix: invalid orb contents, run orb validate in tests, unbound variable

jaredallard:
Fixes the orb to properly use the `run` command syntax (missing the
`run` key.)

Runs `make validate-orb` during the `make test` step to better catch
issues like this in the future.

Fixes an issue where `CI` could be unbound on user's local machines.
Now, we always give it a default value to satisfy `set -u`.


deregtd (approved):
what the label says

kaldorn (approved):

PR #647, opened 2023-08-27
fix: make devspace glibc error

marekfexa-outreach:
- fixes error when glibc version on CircleCI VM (ubuntu latest) does not match with glibc in dev-slim container
- fix is based on disabling [CGO](https://pkg.go.dev/cmd/cgo#:~:text=You%20can%20override%20the%20default,cgo%22%20if%20cgo%20is%20enabled.) which should help with cross-compiling that we do in binary sync (build on macos or ubuntu latest in CircleCI, run in dev-slim container on debian bullseye)
- verified [here](https://app.circleci.com/pipelines/github/getoutreach/devenvtestapp/118/workflows/1ef682dd-04a3-4cd9-aeba-2cc70cb03ea5/jobs/319) that it works

## Jira ID

[QF-936]

## Notes for your reviewers

[QF-936]: https://outreach-io.atlassian.net/browse/QF-936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/Makefile
+++ b/root/Makefile
@@ -183,7 +183,7 @@ devspace-watch:
 .PHONY: devspace
 devspace:
 	@DEVBOX_LOGFMT="$(LOGFMT)" BUILD_FOR_GOOS="linux" CGO_ENABLED=0 SKIP_TRIMPATH="true" SKIP_STARTING_APP="true" DLV_PORT=42097 $(MAGE_CMD) gobuild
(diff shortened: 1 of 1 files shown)


malept (approved):
Seems good to me.

PR #648, opened 2023-09-04
fix(asdf/node): dont install npm again

jaredallard:
Defaults to using the version of `npm` that is present with the
installed version of `node`. This prevents issues where the `npm` latest
version requires a version of node that is not in use, and in turn,
explodes the asdf node default npm packages install step silently.

### DIFF
--- a/shell/ci/env/asdf.sh
+++ b/shell/ci/env/asdf.sh
@@ -36,7 +36,7 @@ init_asdf() {
   git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.12.0
 
   # langauage specifics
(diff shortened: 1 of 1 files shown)


shell/ci/env/asdf.sh
```diff
@@ -36,7 +36,7 @@ init_asdf() {
   git clone https://github.com/asdf-vm/asdf.git ~/.asdf --branch v0.12.0
 
   # langauage specifics
```
malept, line 38: Nit: typo
```suggestion
  # language specifics
```

malept (approved):
Seems fine, just a typo which I will fix and then merge.

malept:
Except it was on automerge :laughing:

PR #650, opened 2023-09-05
fix(orb): ensure proper Node.js version is installed for its gRPC client

malept:

This should fix the issue where the Node.js gRPC client uses a different version of Node.js from the tooling in the root directory, and thus doesn't get installed and fails during the `test_node_client` job.

Also changes references from `node` to `Node.js` where appropriate, to be clearer.

## Notes for reviewers

This will need to be hotfixed.


kaldorn (approved):
Sounds reasonable. This should install node.js.

PR #652, opened 2023-09-06
fix(orb): ensure yarn is installed for test_node_client job

malept:

We've determined that the correct Node.js version is installed via #650, now we ensure that yarn is installed.


malept:
We've figured out that this is not needed.

PR #662, opened 2023-09-07
fix(delibird): fetch delibird token

jaredallard:
Ensures that the delibird CLI has an authentication token configured.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@dev:first
+  shared: getoutreach/shared@2.19.2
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.14.0
+  shared: getoutreach/shared@dev:first
```
malept, line 5: question: why is it using the dev orb?

malept (commented):
This largely looks fine, just a couple of questions before it automerges.

.circleci/config.yml
jaredallard: Ah, leftover from testing this PR on itself

PR #666, opened 2023-09-21
feat: upgrade to yarn v4

jaredallard:
Upgrades to `yarn` v4. Changes the installation method from `npm` to
`corepack`, which is the recommended way to install yarn[^1].

Migrates `devbase` to use v4 of `yarn`, changes some templates to
enable this:
* https://github.com/getoutreach/stencil-base/pull/139
* https://github.com/getoutreach/stencil-golang/pull/375

Added a `asdf_display_versions` function to `lib/asdf.sh` that shows the
current tools and their relevant versions in use. This is displayed during
the `setup_environment` step of CI.

[^1]: https://yarnpkg.com/getting-started/install


jaredallard:
Not sure if I like the asdf versions display I added in this PR, I'll probably make it so it only runs on the setup environment step. Otherwise, it seems like it could pollute CI in other places (see logs)

Makefile
```diff
@@ -18,5 +18,5 @@ validate-orb: build-orb
 
 .PHONY: publish-orb
 publish-orb: validate-orb
-	circleci orb publish orb.yml getoutreach/shared@dev:first
+	circleci orb publish orb.yml getoutreach/shared@dev:jaredallard-feat-yarn-v3
```
jaredallard, line 21: Reset before merge.

shell/languages/nodejs.sh
```diff
@@ -7,13 +7,11 @@ source "$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)/../lib/she
...
-
-  if [[ ! -e "node_modules" ]]; then
-    yarn_install "$stateFile"
-    return
+  if ! command -v yarn >/dev/null; then
+    corepack enable
+    if command -v asdf >/dev/null; then
+      asdf reshim
+    fi
   fi
```
jaredallard, line 15: This is handled by the setup env script, so I'm kinda for just removing this logic in general.

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 # syntax, such as anchors, will be fixed automatically.
 version: 2.1
 orbs:
-  shared: getoutreach/shared@2.19.2
+  shared: getoutreach/shared@dev:jaredallard-feat-yarn-v3
```
jaredallard, line 5: Change back to `dev:first` before merging (publish to `dev:first` before merging locally)

orbs/shared/commands/build_node_client.yml
```diff
@@ -3,4 +3,4 @@ steps:
   - run:
       name: Build Node Client
       working_directory: api/clients/node/
-      command: yarn --frozen-lockfile build
```
jaredallard, line 6: `--frozen-lockfile` isn't required when running scripts now.

orbs/shared/commands/with_node_cache.yaml
```diff
@@ -6,11 +6,11 @@ parameters:
...
   - restore_cache:
       keys:
-        - release-node-cache-{{ checksum "cache-version.txt" }}-{{ checksum "package.json" }}
-        - release-node-cache-{{ checksum "cache-version.txt" }}
+        - node-release-cache-{{ checksum "cache-version.txt" }}-{{ checksum "package.json" }}
+        - node-release-cache-{{ checksum "cache-version.txt" }}
   - run:
       name: Install semantic-release Dependencies
-      command: yarn --frozen-lockfile
+      command: yarn --immutable
```
jaredallard, line 13: `--frozen-lockfile` was replaced by `--immutable`

shell/languages/nodejs.sh
```diff
@@ -7,15 +7,6 @@ source "$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)/../lib/she
 yarn_install_if_needed() {
   local stateFile="node_modules/devbase.lock"
 
-  if ! yarn -v >/dev/null 2>&1; then
```
jaredallard, line 10: Duplicative code, we do this as part of `setup_environment`, which calls `ensure_asdf`, which is ran on local machines and CI.

shell/lib/asdf.sh
```diff
@@ -15,12 +15,18 @@ if [[ -z $asdf_plugins_list ]]; then
   asdf_plugins_list_regenerate
 fi
 
+# asdf_find_tool_versions returns all .tool-version files that could
+# be used in the current repository.
+asdf_find_tool_versions() {
```
jaredallard, line 20: Split out to enable `asdf_display_versions`

shell/linters/prettier.sh
```diff
@@ -11,13 +11,13 @@ extensions=(yaml yml json md ts)
 
 prettier_linter() {
   yarn_install_if_needed >/dev/null
-  find_files_with_extensions "${extensions[@]}" | xargs -n40 "node_modules/.bin/prettier" -l --loglevel log
```
jaredallard, line 14: `node_modules/.bin` isn't a thing in `pnpm` linking. `yarn run <script>` has always supported doing this for us, so this is more clean this way anyways.

PR #672, opened 2023-10-17
fix: support @getoutreach npm packages

jaredallard:
Updates `@getoutreach` scope to always use Github Packages. This is set
globally as our templates rely on Outreach packages. Users that wish to
have their own custom scopes should use a per-repo `.npmrc` instead.

Removes `packagecloud` since it hasn't been supported for a long time
(it also doesn't do anything).

[DT-4053]

[DT-4053]: https://outreach-io.atlassian.net/browse/DT-4053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/ci/auth/github_packages.sh
+++ b/shell/ci/auth/github_packages.sh
@@ -55,5 +55,6 @@ if command -v npm >/dev/null 2>&1; then
   cat >>"$HOME/.npmrc" <<EOF
 
 //npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
+@getoutreach:registry=https://npm.pkg.github.com
 EOF
 fi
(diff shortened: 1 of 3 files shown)


yj-yan (approved):

shell/ci/auth/github_packages.sh
```diff
@@ -55,5 +55,6 @@ if command -v npm >/dev/null 2>&1; then
   cat >>"$HOME/.npmrc" <<EOF
 
 //npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
+@getoutreach:registry=https://npm.pkg.github.com
```
malept, line 58: suggestion: use existing variable
```suggestion
@$ORG:registry=https://npm.pkg.github.com
```

malept (changes requested):
Seems fine but I have one suggestion to make it less annoying to refactor later.

malept (approved):

PR #676, opened 2023-10-18
feat: allow CGO_ENABLED to be configurable for development

malept:

We should not default to compile builds with CGo. This makes it difficult to build a service on a Linux host and run it in a container, particularly if the host and container have different glibc versions. Instead, use `service.yaml` to configure whether CGo compilation is enabled, defaulting to false.

## Jira ID

[DT-4070]

## Notes for your reviewers

A corresponding PR will need to be written for production builds, in `stencil-golang`.

[DT-4070]: https://outreach-io.atlassian.net/browse/DT-4070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shell/cgo-enabled.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# This is a small script to determine whether a service manifest has
+# explicitly enabled compiling with CGo, and print out the appropriate
+# value for the CGO_ENABLED environment variable. Defaults to disabled.
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+# shellcheck source=./lib/bootstrap.sh
+source "$DIR/lib/bootstrap.sh"
+
+if [[ "$(yq -r ".arguments.enableCGo" <"$(get_service_yaml)")" == "true" ]]; then
```
jaredallard, line 10:
```suggestion
if [[ "$(yq -r ".arguments.enableCGO" <"$(get_service_yaml)")" == "true" ]]; then
```

I think it technically should be `CGO`...

jaredallard, line 10: Add field to `manifest.yaml`?

malept: It's unclear to me what the canonical capitalization of it is. It's very inconsistent in https://pkg.go.dev/cmd/cgo.

I guess it's either `Cgo` or `cgo`, so the suggestion should be
```suggestion
if [[ "$(yq -r ".arguments.enableCgo" <"$(get_service_yaml)")" == "true" ]]; then
```

malept: I swear I did that but I'm not seeing it anymore.

jaredallard (approved):

PR #685, opened 2023-11-08
fix(orb): use pre_setup_script in save_cache job

malept:

Despite having the job parameter for it, it wasn't actually being used by the job.


kaldorn (approved):
That does look like it fixes the bug. Approved.

PR #687, opened 2023-11-09
feat: Enable docker layer caching

voslartomas:

Enabling Docker Layer Caching to reduce time of build pipelines. Basically now it downloads and compiles dependencies with every build (even in PRs), this should reduce time of builds when there are no changes in dependencies.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -18,7 +18,7 @@ MANIFEST="$(get_repo_directory)/deployments/docker.yaml"
 
 if [[ -z $TESTING_DO_NOT_BUILD ]]; then
   # shellcheck source=../../lib/buildx.sh
-  source "${LIB_DIR}/buildx.sh"
+  source "${LIB_DIR}/buildx.sh" $APPNAME
 fi
 
 # shellcheck source=../../lib/logging.sh
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -173,7 +173,7 @@ build_and_push_image() {
   fi
   (
     set -x
-    docker buildx build "${args[@]}"
+    docker buildx --builder "$APPNAME" build "${args[@]}"
   )
 }
 
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -173,7 +173,13 @@ build_and_push_image() {
   fi
   (
     set -x
-    docker buildx --builder devbase build "${args[@]}"
+
+    if [[ $OSTYPE == "linux-gnu"* ]]; then
+      docker buildx --builder devbase build "${args[@]}"
+    else
+      docker build "${args[@]}"
+    fi
+
   )
 }
 
--- a/shell/ci/release/docker.sh
+++ b/shell/ci/release/docker.sh
@@ -18,7 +18,7 @@ MANIFEST="$(get_repo_directory)/deployments/docker.yaml"
 
 if [[ -z $TESTING_DO_NOT_BUILD ]]; then
   # shellcheck source=../../lib/buildx.sh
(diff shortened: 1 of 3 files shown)


shell/lib/buildx.sh
```diff
@@ -41,7 +42,7 @@ if [[ $OSTYPE == "linux-gnu"* ]]; then
 
   # Take from setup-buildx Github Action
   echo "� Creating a new builder instance"
-  docker buildx create --use
+  docker buildx create --use --name "$APPNAME"
```
jaredallard, line 45: This is also ran on user's machines, so perhaps we utilize a shared one instead of one per-application? Also, we probably need to `|| true` this or something smarter to prevent this from failing.

I would probably just go with `--name devbase` and share it everywhere.

shell/lib/buildx.sh
```diff
@@ -5,6 +5,7 @@
 # BUILDX_VERSION is the version of 'docker buildx' we should use if not
 # already installed on the host system.
 BUILDX_VERSION="v0.11.2"
+APPNAME="$1"
```
jaredallard, line 8: If we do keep this, please document it using the format above, example:

```
# APPNAME is xyz and is used for the buildx builder's name.
```

This helps with development because there are things like [bash lsp](https://github.com/bash-lsp/bash-language-server) that can read this and make it available on hover, for example.

jaredallard:
Also, please fill out the PR description with the motivations behind this change (optionally a link to the slack context, but since this is internal it might be best to not).

shell/lib/buildx.sh
voslartomas: Ah yeah right, I guess cache is per repository so we don't need such unique names.

shell/lib/buildx.sh
voslartomas: Simplified, no need for comment anymore

shell/ci/release/docker.sh
```diff
@@ -173,7 +173,7 @@ build_and_push_image() {
   fi
   (
     set -x
-    docker buildx build "${args[@]}"
+    docker buildx --builder devbase build "${args[@]}"
```
jaredallard, line 176: I just realized L34 of `buildx.sh` scripts the builder creation on macOS (because it isn't required). we'd probably have to encapsulate that logic here as well (omit the flag on macOS).

jaredallard: I would just conditionally append to the `args` associative array.

jaredallard: (or, if positional, make a new array to pass to it, to keep `set -x` from being noisy)

shell/ci/release/docker.sh
```diff
@@ -173,7 +173,13 @@ build_and_push_image() {
   fi
   (
     set -x
```
jaredallard, line 175: Can you remove the `set -x`, once there's `if` statements in here, it just makes it super complicated to read the output 😓 

```suggestion
```

shell/ci/release/docker.sh
```diff
@@ -173,7 +173,13 @@ build_and_push_image() {
...
-    docker buildx build "${args[@]}"
+
+    if [[ $OSTYPE == "linux-gnu"* ]]; then
+      docker buildx --builder devbase build "${args[@]}"
+    else
+      docker build "${args[@]}"
```
jaredallard, line 180: Still need `buildx`, just not `--builder`.

```suggestion
      docker buildx build "${args[@]}"
```

voslartomas: Ah I was trusting copilot too much this time

jaredallard (approved):

PR #690, opened 2023-11-10
feat: Tweak cache

voslartomas:

Docker cache keeps increasing with each build and will reset itself once it hits 15G in circleci, this will prune cache after each build and will keep usage under 6GB. 

Another solution could be to switch to different provider, rather then DLC in circleci. For example https://docs.docker.com/build/cache/backends/s3/

## Jira ID

[XX-XX]


jaredallard:
I've had mixed results with using docker buildx cache backends (on Github Actions, anyways). The prune change here makes sense though, I like it.

jaredallard (approved):

PR #691, opened 2023-11-10
feat(orb): add jobs to build and stitch docker images

malept:

Replaces the existing job which did all of the image building and pushing (which includes building the arm64 image in qemu), with three jobs:

* build amd64 image
* build arm64 image
* put the two images in the same manifest and push to remote registry

It is intended that the first two jobs are required by the third job.

## Jira ID

[DT-3606]

## Notes for your reviewers

It completely replaces the old job so there's less chance of it conflicting when `devbase` (and `stencil-circleci`) is upgraded.

The daily cache key needed to be augmented to have the arch in it, so that the `docker_arm64` job didn't fail by running cached amd64 binaries.

[DT-3606]: https://outreach-io.atlassian.net/browse/DT-3606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kaldorn (approved):
We synced. This all makes sense.

PR #698, opened 2023-11-18
fix(ci): handle the case where multiple images are defined for a given service

malept:

Does as it says on the tin.

## Jira ID

[DT-3606]

## Notes for your reviewers

I moved `get_image_field` into its own library file, so I moved its corresponding tests plus got rid of the `TESTING_DO_NOT_BUILD` env var hack.

I also extracted the logic to generate `docker buildx` args into its own function and added tests.

[DT-3606]: https://outreach-io.atlassian.net/browse/DT-3606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/docker.sh
+++ b/shell/lib/docker.sh
@@ -42,3 +42,11 @@ get_image_field() {
     fatal "Expected 'array' or 'string'"
   fi
 }
+
+# run_docker is a wrapper for the docker command, but it prints out the
+# command (via set -x).
+run_docker() {
+  set -x
+  docker "$@"
+  set +x
+}
--- a/shell/lib/docker.sh
+++ b/shell/lib/docker.sh
@@ -6,6 +6,9 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 # shellcheck source=./logging.sh
 source "${DIR}/logging.sh"
 
+# shellcheck source=./bootstrap.sh
+source "${DIR}/bootstrap.sh"
+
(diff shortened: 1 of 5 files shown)


shell/lib/docker.sh
```diff
@@ -0,0 +1,174 @@
...
+  # Platforms to build this image for, expected format (in YAML):
+  # platforms:
+  #   - linux/arm64
+  #   - linux/amd64
+  #
+  # See buildkit docs: https://github.com/docker/buildx#building-multi-platform-images
+  mapfile -t platforms < <(get_image_field "$image" "platforms" "array")
+  if [[ ${#platforms[@]} -eq 0 ]]; then # no platforms found
+    if [[ -n $arch ]]; then
+      platforms=("linux/${arch}")
```
yj-yan, line 68: nit: If the $arch is an input from yaml, suggest to validate and only allow for the predefined platforms.

yj-yan (approved):
I am not an bash expert. The logic seems fine and the tests look good.

shell/lib/docker.sh
malept: I don't *particularly* want to try to validate the value here, since this is an open source repo and external users could use arches that are outside of `amd64`/`arm64` (this [Docker documentation page](https://docs.docker.com/build/building/multi-platform/) has a non-exhaustive list of supported arches).

kaldorn (approved):
This seems reasonable to me.

PR #704, opened 2023-12-05
fix: ensure asdf utilities are run with bash 5 or greater

malept:

the `asdf` utilities file uses `readarray`, which is not implemented in the version of bash that comes with macOS (which is 3.2). So, check the bash version array's major version at the beginning of the file.

Tested on macOS (thanks @jaredallard) and Linux with both Bash 3.2 and Bash 5.x.

## Jira ID

[DT-4123]

[DT-4123]: https://outreach-io.atlassian.net/browse/DT-4123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/ci/testing/delibird.sh
+++ b/shell/ci/testing/delibird.sh
@@ -7,15 +7,18 @@
 set -euo pipefail
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+LIB_DIR="$DIR/../../lib"
+
+echo "LIB DIR: $LIB_DIR"
 
 # shellcheck source=../../lib/bootstrap.sh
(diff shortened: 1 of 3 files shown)


jaredallard (approved):
I would like this to be moved somewhere like `shell.sh` longer-term so we can detect this in all scripts automatically, but we don't really have a good entrypoint like that today, so I'm fine with this as-is!

malept:
> I would like this to be moved somewhere like `shell.sh` longer-term so we can detect this in all scripts automatically, but we don't really have a good entrypoint like that today, so I'm fine with this as-is!

I ended up doing that.

PR #714, opened 2023-12-12
fix(box): abort early if the box repo can't be cloned correctly

malept:

In addition to what it says on the tin, the PR fixes the following:

* Delete the generated temporary directory instead of the entire system `TMPDIR` on `EXIT` signal captured
* Put `yq` flag in the correct position

## Jira ID

[DT-4111]

[DT-4111]: https://outreach-io.atlassian.net/browse/DT-4111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/box.sh
+++ b/shell/lib/box.sh
@@ -42,7 +42,7 @@ download_box() {
   local newBox=$(yq . "${BOXPATH}" |
     jq --slurpfile boxconf \
       <(yq -r . "${tempDir}/box.yaml") '. * {config: $boxconf[0] }' |
-    yq . --yaml-output)
+    yq --yaml-output .)
   echo "${newBox}" >"${BOXPATH}"
 }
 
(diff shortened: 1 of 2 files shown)


shell/lib/box.sh
```diff
@@ -27,10 +27,15 @@ download_box() {
   # Why: OK with assigning without checking exit code.
   # shellcheck disable=SC2155
   local tempDir="$(mktemp -d)"
-  trap 'rm -rf "${TMPDIR}"' EXIT
+  trap 'rm -rf "${tempDir}"' EXIT
```
jaredallard, line 30: oops

jaredallard (approved):

yj-yan (approved):

PR #716, opened 2023-12-13
fix: use CGO_ENABLED for make gobuild

malept:

This is for consistency with the other commands which utilize `CGO_ENABLED`.


jaredallard (approved):
My approval means nothing now, but I APPROVE

yj-yan (approved):

PR #721, opened 2024-01-10
feat: make go test timeout configurable in test job

kaldorn:

Unfortunately, I have a test that takes longer than 10 minutes and this isn't currently configurable.

The test.sh looks for this env var and sets the timeout flag:
https://github.com/getoutreach/devbase/blob/be1b62d86f8295ea542f9adecdf7ace7990cf5bc/shell/test.sh#L103-L106

## Jira ID

[XX-XX]

### DIFF
--- a/orbs/shared/jobs/test.yaml
+++ b/orbs/shared/jobs/test.yaml
@@ -42,6 +42,10 @@ parameters:
   setup_remote_docker:
     type: boolean
     default: false
+  go_test_timeout:
+    description: Maps to gotest -timeout parameter.
+    type: string
+    default: ""
 
 resource_class: << parameters.resource_class >>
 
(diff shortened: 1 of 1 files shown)


orbs/shared/jobs/test.yaml
```diff
@@ -53,6 +57,7 @@ executor:
 environment:
   AWS_REGION: << parameters.aws_region >>
   PRE_SETUP_SCRIPT: << parameters.pre_setup_script >>
+  GO_TEST_TIMEOUT: << parameters.go_test_timeout >>
```
george-e-shaw-iv, line 60: so i take it this is used in a script somewhere when we run the tests?

kaldorn: Yup.
https://github.com/getoutreach/devbase/blob/be1b62d86f8295ea542f9adecdf7ace7990cf5bc/shell/test.sh#L103-L106

malept (approved):
This looks like it's the same as the one for the E2E tests, so :shipit:

PR #738, opened 2024-02-29
fix: replace yq with gojq wrapper

### LINKED ISSUE #457: `yq` is a hard requirement for `make`
Running `make` on a fresh build requires `yq` to even pull down `devbase`. There is an `asdf` plugin for `yq`, but it installs the Go `yq` and not the Python `yq` (and I believe that the Python `yq` is expected). 

In the short term, this hard dependency should be documented. For the long term, should this be an be an `asdf` plugin of its own, or should the project move to the Go `yq` which is much more portable and doesn't require Python.

malept:

Replaces all instances of `yq` with a wrapper script that will download and run `gojq` (which has YAML support) if `python-yq` isn't installed. Currently if prefers `python-yq`, but this will change in the future. There is an environment variable to force the use of `gojq`, `YQ_USE_GOJQ`.

In `scripts/devbase.sh`, which bootstraps `devbase` in a given stenciled repository, it just downloads `gojq` to a cached directory and uses that. The overhead on the first run on my laptop is ~2.5 seconds and effectively nothing after that.

Fixes #457.


root/Makefile
```diff
@@ -1,6 +1,7 @@
 SHELL                := /usr/bin/env bash
 
-BOOTSTRAP_DIR        := $(shell if [[ "$$(yq -r .name service.yaml 2>/dev/null || echo "")" == "devbase" ]]; then echo "$(CURDIR)" ; else echo "$(CURDIR)/.bootstrap"; fi)
+YQ                   := $(CURDIR)/.bootstrap/shell/yq.sh
+BOOTSTRAP_DIR        := $(shell if [[ "$$("$(YQ)" -r .name service.yaml 2>/dev/null || echo "")" == "devbase" ]]; then echo "$(CURDIR)" ; else echo "$(CURDIR)/.bootstrap"; fi)
```
jaredallard, line 4: Good job me

scripts/devbase.sh
```diff
@@ -6,25 +6,42 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
...
+  local field="$1"
+  local file="$2"
 
-  if [[ "$(yq e '.a' '-' <<<'{"a": "true"}' 2>&1)" == "true" ]]; then
-    # using golang version
-    yq e "$field" "$file"
-  else
-    # probably using python version
-    yq -r "$field" <"$file"
+  local gjDir
```
jaredallard, line 22: **thought:** Probably best to compose this logic into a separate function and call out to it? Perhaps run it whenever this library is imported (sourced)?

shell/circleci/machine.sh
```diff
@@ -47,24 +47,12 @@ if ! command -v gh >/dev/null; then
   rm ./gh.deb
 fi
 
-echo "Installing yq"
+echo "Removing yq"
 # Remove the existing yq, if it already exists
 # (usually the Go Version we don't support)
+# We can use gojq via a yq wrapper which uses gobin.sh
 sudo rm -f "$(command -v yq)"
```
jaredallard, line 54: **thought:** Hah, YOLO nuke `yq`. I like it.

You probably don't need to nuke whatever exists given your wrapper, though.

shell/lib/bootstrap.sh
jaredallard, line 1: **thought:** It's probably a good idea to change this file to call the `yaml.sh` functions instead someday.

shell/lib/box.sh
```diff
@@ -44,10 +46,10 @@ download_box() {
 
   # Why: OK with assigning without checking exit code.
   # shellcheck disable=SC2155
-  local newBox=$(yq . "${BOXPATH}" |
+  local newBox=$("$YQ" . "${BOXPATH}" |
     jq --slurpfile boxconf \
```
jaredallard, line 50: **thought:** Could probably remove the `jq` dependency here at some point also, right?

shell/lib/docker.sh
```diff
@@ -172,3 +173,10 @@ run_docker() {
   docker "$@"
   set +x
 }
+
+# docker_manifest_images retrieves the list of images from a manifest file.
+docker_manifest_images() {
+  local manifest="$1"
+
+  "$YQ" -r 'keys[]' "$manifest"
```
jaredallard, line 181: **nitpick:** Not much value to using a local variable for a one-liner is there?

shell/yq.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+#
+# Wrapper around `yq` that allows use of python-yq or gojq.
+# You can force using gojq by setting the environment variable YQ_USE_GOJQ=true.
+#
+
```
jaredallard, line 6: Probably want to use `set -euo pipefail` in all new shell files.

shell/yq.sh
```diff
@@ -0,0 +1,28 @@
...
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
+if [[ -z $GOJQ_VERSION ]]; then
+  GOJQ_VERSION="$(grep ^gojq: "$DIR"/../versions.yaml | awk '{print $2}')"
+fi
```
jaredallard, line 11:
```suggestion
GOJQ_VERSION="${GOJQ_VERSION:-$(grep ^gojq: "$DIR"/../versions.yaml | awk '{print $2}')}"
```

Easier to read for people who write a lot of bash. But also, I hate this fake yaml parsing you're doing here.

shell/yq.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [[ -n $YQ_USE_GOJQ ]]; then
+  use_gojq=true
+elif command -v yq >/dev/null 2>&1; then
+  # Make sure it's the correct yq. The Go yq is not compatible jq syntax.
```
jaredallard, line 16: This branch of code is going to leave `use_gojq` to an empty value, right? That feels a little weird. Maybe default to `false` or something else?

templates/scripts/devbase.sh.tpl
jaredallard, line 1: **thought:** This makes me cry inside. Duplication :cry:

jaredallard (commented):
Here's a completely unsolicited review :ghost:

shell/lib/box.sh
malept: I would need to figure out what the heck `--slurpfile` does first :sweat_smile:

shell/lib/docker.sh
malept: If bash function args were named, sure.

shell/yq.sh
malept: Fair
```suggestion

set -euo pipefail
```

shell/yq.sh
malept: > Easier to read for people who write a lot of bash.

I think I want to optimize for folks who don't read a lot of bash at this point, given the current number of reviewers with bash experience.

> I hate this fake yaml parsing you're doing here.

Well, there's a bootstrapping problem. Can't use `yq`/`gojq` on a YAML file if we don't have it installed yet.

shell/yq.sh
malept: Good catch, will fix.

templates/scripts/devbase.sh.tpl
malept: I mean, one's a template and one's the rendered file via `stencil`. Not sure how to fix this without some thought into redesigning.

shell/circleci/machine.sh
malept: Yeah upon reflection, I don't think we can actually do this change right now since there's a bunch of other scripts which depend on `yq` at the moment. Maybe later if I can force install a `yq` wrapper.

shell/yq.sh
```diff
@@ -0,0 +1,30 @@
...
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
+GOJQ_VERSION="${GOJQ_VERSION:-$(grep ^gojq: "$DIR"/../versions.yaml | awk '{print $2}')}"
+
+use_gojq=false
+
+if [[ -n $YQ_USE_GOJQ ]]; then
+  use_gojq=true
+elif command -v yq >/dev/null 2>&1; then
```
yj-yan, line 19:
```suggestion
 # Make sure it's the correct yq. The Go yq is not compatible with jq syntax.
elif ! (command -v yq >/dev/null 2>&1 && [[ "$(yq --version)" =~ ^yq.3.* ]]);  then
  use_gojq=true
```
Suggest: could make this conditional check simplified. (Github does not allow to choose multiple lines to suggest somehow. )

yj-yan (commented):
Probably need to resolve the install path for MacOS. I tried stencil the branch and got the error:
```
tar: Error opening archive: Failed to open '/Users/yyan/.cache/devbase/gojq/gojq_v0.12.14_darwin_arm64.tar.gz'
mv: rename /Users/yyan/.cache/devbase/gojq/gojq to /Users/yyan/.cache/devbase/gojq/gojq-v0.12.14: No such file or directory
./scripts/devbase.sh: line 44: /Users/yyan/.cache/devbase/gojq/gojq-v0.12.14: No such file or directory
```

shell/yq.sh
malept: I'm not sure if that conditional is actually simpler, since you're negating a predicate which is already pretty complicated - it's a mess of symbols. I'd rather be more verbose when possible with bash for this reason.

yj-yan: That's fair.

yj-yan (approved):

malept:
> > Probably need to resolve the install path for MacOS. I tried stencil the branch and got the error:
> > ```
> > tar: Error opening archive: Failed to open '/Users/yyan/.cache/devbase/gojq/gojq_v0.12.14_darwin_arm64.tar.gz'
> > mv: rename /Users/yyan/.cache/devbase/gojq/gojq to /Users/yyan/.cache/devbase/gojq/gojq-v0.12.14: No such file or directory
> > ./scripts/devbase.sh: line 44: /Users/yyan/.cache/devbase/gojq/gojq-v0.12.14: No such file or directory
> > ```
> 
> We have tested on MacOS and it works now.

As it turns out, the Linux binary is in a `.tar.gz` whereas the macOS and Windows binaries are in a `.zip`. :rage:

PR #746, opened 2024-03-08
feat(golangci-lint): Add GOMEMLIMIT to env

Ned-el-ch:

Adds GOMEMLIMIT with a default of 6GiB to the e2e env variables

## Jira ID

[DT-4162]

## Notes for your reviewers

[DT-4162]: https://outreach-io.atlassian.net/browse/DT-4162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pbr0ck3r (approved):

malept:
> We may want to document somewhere the different CirceCI image containers and what their max memory limit is. If a team increases container size to say xlarge to speed things up, they will want to bump the `gomemlimit` to take advantage.

Should this be part of the description?

yj-yan (approved):

malept (changes requested):
I'm confused. The original issue was in the `shared/test` job, so why is the fix in the `shared/e2e` job?

Ned-el-ch:
> I'm confused. The original issue was in the `shared/test` job, so why is the fix in the `shared/e2e` job?

the jira ticket said e2e so it went there but i'll change it 
> We could also also the GOMEMLIMIT in .circleci/config in the tests (e2e) steps, as it can't be set in a workflow via stencil-circleci. [Set an environment variable - CircleCI](https://circleci.com/docs/set-environment-variable/)

orbs/shared/jobs/test.yaml
```diff
@@ -58,6 +62,7 @@ environment:
   AWS_REGION: << parameters.aws_region >>
   PRE_SETUP_SCRIPT: << parameters.pre_setup_script >>
   GO_TEST_TIMEOUT: << parameters.go_test_timeout >>
+  GOMEMLIMIT: << parameters.gomemlimit >>
```
malept, line 65: issue: when you set this, the entire check for heuristically determining the value of `GOMEMLIMIT` is always bypassed due to this check:

https://github.com/getoutreach/devbase/blob/239fd4372b6e3d360cb351c000e96958b66025ef/shell/golangci-lint.sh#L44

We need to figure out a better way of allowing repos to specify the value without removing the heuristic calculations feature.

shell/golangci-lint.sh
```diff
@@ -53,7 +53,9 @@ if [[ -z $GOGC ]] && [[ -z $GOMEMLIMIT ]]; then
   # This is mostly important for CI systems or other container
   # environments.
   if (command -v free || command -v sysctl) &>/dev/null; then
-    if command -v free &>/dev/null; then
+    if [ -f /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
+      mem=$(($(cat /sys/fs/cgroup/memory/memory.limit_in_bytes) / 1024 / 1024))
```
malept, line 57: suggestion: it occurred to me that this doesn't seem to be related to `free` or `sysctl` being available, so this `if` statement could be on the same level as line 55, instead of being nested inside of line 55.

PR #763, opened 2024-04-11
fix(gojq): allow fetching via busybox or wget

malept:

There is a use case where `make gogenerate` (and thus, `scripts/devbase.sh`) are run in an Alpine Linux Docker container. In that case, `curl` doesn't exist by default, but `busybox` does, which has a `wget` applet. If `busybox` is detected, prefer using that over `curl`.

Also add support for GNU `wget` since it's effectively the same command as the busybox one.


pbr0ck3r (approved):

yj-yan (approved):
Alternative download tools, LGTM.

PR #771, opened 2024-04-16
feat: add `make stencil` and `make restencil`

malept:

Also fix restenciling `devbase` due to some of the special things that this stencil module needs.

This is needed for some Cortex work.

## Jira ID

[DT-4321]

[DT-4321]: https://outreach-io.atlassian.net/browse/DT-4321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


root/Makefile
```diff
@@ -230,6 +232,22 @@ docs:: pre-docs
...
+## stencil:         run stencil
+.PHONY: stencil
+stencil:: pre-stencil run-stencil post-stencil
+
+.PHONY: run-stencil
+run-stencil:
+	stencil --skip-update
+
+## restencil:       run stencil without updating module versions
+.PHONY: restencil
```
yj-yan, line 244: Is this `restencil` only applied to the devbase module?

malept: Since this is in `root/Makefile` and not `Makefile`, this is applied to all stenciled repos using this version.

yj-yan: I see. I take this is as the special case without updating devbase module, but it is no harm to add the feature for all stenciled repos.

yj-yan (approved):

root/Makefile
malept: `make restencil` is the equivalent of someone running `stencil --frozen-lockfile`, for example if they change the `reportingTeam` in `service.yaml` and don't want to upgrade any stencil modules at the same time.

PR #774, opened 2024-04-17
feat: add catalog sync script, used when catalog manifest isn't a template

malept:

Does what it says on the tin. This script only updates some basic fields, it's not exhaustive. An exhaustive updater would effectively require a Go-based script. It currently only supports Cortex but can be extended to other service catalogs' manifests.

## Jira ID

[DT-4324]

[DT-4324]: https://outreach-io.atlassian.net/browse/DT-4324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 APP := devbase
 OSS := true
-_ := $(shell ./scripts/devbase.sh)
+_ := $(shell ./scripts/devbase.sh) 
 
 include .bootstrap/root/Makefile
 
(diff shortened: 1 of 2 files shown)


yj-yan (approved):
Shell script to sync from various manifest files to cortex.yaml, looks legit.

PR #808, opened 2024-08-16
feat: add configurable push container registries

malept:

Allows for pushing to multiple container registries.

Also adds support to authenticate with AWS ECR.

## Jira ID

[DT-4464]

[DT-4464]: https://outreach-io.atlassian.net/browse/DT-4464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ekalinichev (approved):
Looks great, job well done!

As a side-note -- amout of code is close to uncomfortable to use without tests, we'll have to move from Shell to something more controllable at some point in future

malept:
> As a side-note -- amout of code is close to uncomfortable to use without tests, we'll have to move from Shell to something more controllable at some point in future

There is infrastructure to start migrating everything to Go (via Mage) but it hasn't yet been a priority.

That being said I can add some tests for the new `get_box_array` function.

PR #846, opened 2024-12-20
fix(get_app_version): work as intended, add tests

malept:

Porting over the Magefile version of `get_app_version` did not work as intended, and it's unclear why.

* Removed `--always` from `git describe` so that a repo without any tags actually uses the fallback.
* Updated the value passed to `--match` to be a glob instead of a regex, per the docs.
* Minor change: prefix fallback version with `v` for consistency

Also adds tests to show what it's supposed to do.

## Jira ID

[DT-4561]

## Notes for your reviewers

`--always` documentation from `git describe --help`:

```
       --always
           Show uniquely abbreviated commit object as fallback.
```

[DT-4561]: https://outreach-io.atlassian.net/browse/DT-4561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/bootstrap.sh
+++ b/shell/lib/bootstrap.sh
@@ -39,7 +39,7 @@ get_app_name() {
 # get_app_version returns the version of the application
 # from git.
 get_app_version() {
-  git describe --match='v[0-9]+' --tags --always HEAD 2>/dev/null || echo "0.0.0-dev"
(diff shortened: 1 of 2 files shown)


richardlarocque (approved):
LGTM.

ekalinichev (approved):
There are unit tests for Bash? Nice )

PR #879, opened 2025-03-18
feat: replace gcr.io links with ecr link

cyrus-smith-outreach:

Giraffe deployment [is not working](https://outreach-hq.slack.com/archives/CN9MU7GLW/p1742239257335989). I believe this is because the devbase local deployment scripts are using `gcr.io` as the image registry. Giraffe (and likely other services have already switched to use ecr so the images are not available on ecr.

## Jira ID

[XX-XX]

### DIFF
--- a/shell/build-jsonnet.sh
+++ b/shell/build-jsonnet.sh
@@ -18,7 +18,7 @@ version="${DEVENV_DEPLOY_VERSION:-"latest"}"
 environment="${DEVENV_DEPLOY_ENVIRONMENT:-"development"}"
 host="${DEVENV_DEPLOY_HOST:-"bento1a.outreach-dev.com"}"
 email="${DEV_EMAIL:-$(git config user.email)}"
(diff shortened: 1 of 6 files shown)


shell/build-jsonnet.sh
```diff
@@ -18,7 +18,7 @@ version="${DEVENV_DEPLOY_VERSION:-"latest"}"
 environment="${DEVENV_DEPLOY_ENVIRONMENT:-"development"}"
 host="${DEVENV_DEPLOY_HOST:-"bento1a.outreach-dev.com"}"
 email="${DEV_EMAIL:-$(git config user.email)}"
-appImageRegistry="${DEVENV_DEPLOY_IMAGE_REGISTRY:-"gcr.io/outreach-docker"}"
```
cyrus-smith-outreach, line 21: I think this is the change that should fix devenv apps deploy.

I have not tested this locally and I'm not sure how I can.

PR #895, opened 2025-04-29
fix: replace tool installs with mise

malept:

Does what it says on the tin.

## Jira ID

[DT-4640]

## Notes for your reviewers

Drops JIT install of `gojq` in `shell/yq.sh` - it should be installed everywhere at this point.

[DT-4640]: https://outreach-io.atlassian.net/browse/DT-4640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shell/ci/testing/setup-devenv.sh
```diff
@@ -10,6 +10,8 @@ source "$DIR/../../lib/logging.sh"
 source "$DIR/../../lib/github.sh"
 # shellcheck source=../../lib/box.sh
 source "$DIR/../../lib/box.sh"
+# shellcheck source=../../lib/mise.sh
```
iamrz1, line 13: do you plan to keep this for future reference?

malept: @iamrz1 yes, why do you ask?

iamrz1 (approved):

PR #896, opened 2025-05-01
fix(linters): allow prettier to be installed via `mise` instead of `package.json`

malept:

This is for internal Stenciled non-service repos which don't need to use `stencil-base` (which provides `prettier` in its templated `package.json`). Instead, it assumes that the repo has installed it via `mise`.

### DIFF
--- a/shell/linters/prettier.sh
+++ b/shell/linters/prettier.sh
@@ -10,6 +10,13 @@ source "$DIR/languages/nodejs.sh"
 extensions=(yaml yml json md ts)
 
 PRETTIER="node_modules/.bin/prettier"
+if [[ ! -f "$PRETTIER" ]]; then
+  # Try to find prettier installed via mise
+  PRETTIER="$(mise which prettier)"
+  if [[ -z $PRETTIER ]]; then
+    fatal "prettier not found in repo"
+  fi
+fi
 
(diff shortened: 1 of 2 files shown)


shell/linters/prettier.sh
```diff
@@ -10,6 +10,13 @@ source "$DIR/languages/nodejs.sh"
 extensions=(yaml yml json md ts)
 
 PRETTIER="node_modules/.bin/prettier"
+if [[ ! -f $PRETTIER ]] && [[ ! -f package.json ]]; then
+  # Try to find prettier installed via mise
+  PRETTIER="$(mise which prettier)"
+  if [[ -z $PRETTIER ]]; then
+    fatal "prettier not found in repo"
```
ekalinichev, line 17: Should we update wording of the message? Current one will cause people to write to #dev-tooling-support, maybe we can prompt them to install prettier via `mise`?

ekalinichev (approved):

shell/linters/prettier.sh
malept:
```suggestion
    fatal "prettier not found in repo, make sure 'npx:prettier' is defined in 'mise.toml' and you have run 'mise install'"
```

PR #906, opened 2025-05-20
fix: cache jsonnet-libs to avoid 429 errors

lelandbatey:

Currently, we're seeing many services failing their E2E tests in CI because we include a public repo that we maintain called `jsonnet-libs` via a `raw.githubusercontent.com` URL, so jsonnet files imported by `kubecfg` are imported directly over HTTPS. This has predictably led to us getting rate-limited, as our tooling can end up making a ton of requests to that repo specifically, and Github apparently limits requests to `raw.githubusercontent.com` to ~5000/hour/ip address.

https://github.com/github/docs/issues/8031#issuecomment-880564549

To avoid making so many HTTPS request, this modification to the `build-jsonnet.sh` script clones a local copy of the `jsonnet-libs` directory to the local disk, placing that locally cached copy in the `$HOME/.outreach/.cache` folder as that's the blessed path for caching data for tooling (other scripts do the same e.g. `shell/lib/shell.sh`).

The estimated API usage reduction is +10x since before we'd make 1 request per file (there's 15+ jsonnet files in `jsonnet-libs`), now we clone at most once per run. There are still limits on clones (~5000/hour/authed user), but with our reduced usage we're unlikely to hit that limit.

## Jira ID

No known JIRA for this.

### DIFF
--- a/shell/build-jsonnet.sh
+++ b/shell/build-jsonnet.sh
@@ -11,17 +11,19 @@ source "$SCRIPTS_DIR/lib/docker.sh"
 # shellcheck source=./lib/logging.sh
 source "$SCRIPTS_DIR/lib/logging.sh"
 
-# Cache a local copy of the `jsonnet-libs` directory on disk if it doesn't yet exist. We use the SHA
-# of the latest HEAD for invalidation; as HEAD changes with new commits, so will our cache. Do this
+# Cache a local copy of the `jsonnet-libs` directory on disk if it doesn't yet exist. Do this
 # because it helps us avoid accessing jsonnet-libs via raw.githubusercontent.com, which has
 # aggressive rate limits that we can easily hit. Estimated API usage reduction is +10x since before
 # we'd make 1 request per file (15+ *sonnet files), now we clone at most once per run.
-JSONNET_LIBS_SHA="$(git ls-remote 'https://github.com/getoutreach/jsonnet-libs.git' HEAD | cut -f1)"
-JSONNET_LIBS_CACHE="$HOME/.outreach/.cache/jsonnet-libs/${JSONNET_LIBS_SHA}"
+JSONNET_LIBS_REPO="$HOME/.outreach/.cache/jsonnet-libs"
 
-mkdir -p "$(dirname "$JSONNET_LIBS_CACHE")"
-if [[ ! -f "$JSONNET_LIBS_CACHE" ]]; then
-  git clone -q --single-branch git@github.com:getoutreach/jsonnet-libs "$JSONNET_LIBS_CACHE" >/dev/null
+if [[ -d "$JSONNET_LIBS_REPO" ]]; then
+  pushd "$JSONNET_LIBS_REPO" || fatal "Could not find jsonnet-libs cache dir"
+  git pull
+  popd || fatal "Could not change directory out of jsonnet-libs cache dir"
+else
+  mkdir -p "$(dirname "$JSONNET_LIBS_REPO)"
+  git clone --quiet --single-branch git@github.com:getoutreach/jsonnet-libs "$JSONNET_LIBS_CACHE" >/dev/null
 fi
 
 action=$1
--- a/shell/build-jsonnet.sh
+++ b/shell/build-jsonnet.sh
@@ -11,6 +11,19 @@ source "$SCRIPTS_DIR/lib/docker.sh"
 # shellcheck source=./lib/logging.sh
 source "$SCRIPTS_DIR/lib/logging.sh"
 
+# Cache a local copy of the `jsonnet-libs` directory on disk if it doesn't yet exist. We use the SHA
+# of the latest HEAD for invalidation; as HEAD changes with new commits, so will our cache. Do this
+# because it helps us avoid accessing jsonnet-libs via raw.githubusercontent.com, which has
+# aggressive rate limits that we can easily hit. Estimated API usage reduction is +10x since before
+# we'd make 1 request per file (15+ *sonnet files), now we clone at most once per run.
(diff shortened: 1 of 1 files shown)


shell/build-jsonnet.sh
```diff
@@ -11,6 +11,19 @@ source "$SCRIPTS_DIR/lib/docker.sh"
 # shellcheck source=./lib/logging.sh
 source "$SCRIPTS_DIR/lib/logging.sh"
 
+# Cache a local copy of the `jsonnet-libs` directory on disk if it doesn't yet exist. We use the SHA
+# of the latest HEAD for invalidation; as HEAD changes with new commits, so will our cache. Do this
+# because it helps us avoid accessing jsonnet-libs via raw.githubusercontent.com, which has
+# aggressive rate limits that we can easily hit. Estimated API usage reduction is +10x since before
+# we'd make 1 request per file (15+ *sonnet files), now we clone at most once per run.
+JSONNET_LIBS_SHA="$(git ls-remote 'https://github.com/getoutreach/jsonnet-libs.git' HEAD | cut -f1)"
+JSONNET_LIBS_CACHE="$HOME/.outreach/.cache/jsonnet-libs/${JSONNET_LIBS_SHA}"
```
AndrewWinterman, line 20: Would this support other directories too? I could start using this approach for Redis, for example

lelandbatey: As far as I know, anything in `devbase` is free to stuff whatever they like into `$HOME/.outreach/.cache/` as long as there's no stepping on toes. If you're doing stuff _outside_ devbase, I have no idea if that's _allowed_ or not per-se, but I don't think anyone would mind if you cached stuff in `$HOME/.outreach/.cache/` from whatever tooling codebase you have in mind.

dvagapov (approved):

shell/build-jsonnet.sh
```diff
@@ -11,6 +11,19 @@ source "$SCRIPTS_DIR/lib/docker.sh"
...
+
+mkdir -p "$(dirname "$JSONNET_LIBS_CACHE")"
+if [[ ! -f "$JSONNET_LIBS_CACHE" ]]; then
+  git clone -q --single-branch git@github.com:getoutreach/jsonnet-libs "$JSONNET_LIBS_CACHE" >/dev/null
+fi
```
malept, line 25: suggestion: I feel like this is going to clone a new `jsonnet-libs` every time there's a new commit. Why not just have one directory that gets updated via git pull?
```suggestion
JSONNET_LIBS_REPO="$HOME/.outreach/.cache/jsonnet-libs"

if [[ -d "$JSONNET_LIBS_REPO" ]]; then
  pushd "$JSONNET_LIBS_REPO" || fatal "Could not find jsonnet-libs cache dir"
  git pull
  popd || fatal "Could not change directory out of jsonnet-libs cache dir"
else
  mkdir -p "$(dirname "$JSONNET_LIBS_REPO)"
  git clone --quiet --single-branch git@github.com:getoutreach/jsonnet-libs "$JSONNET_LIBS_CACHE" >/dev/null
fi
```

malept (changes requested):

shell/build-jsonnet.sh
lelandbatey: That seems a good choice to me; the "cache a copy of each commit" would only potentially be useful if we where somehow tracking and pinning the version of the `jsonnet-libs` in each repo, but we're not doing that and we never have done that, we've only ever used tip-of-main via the HTTP import. I like this, it'll keep things cleaner on disk.

shell/build-jsonnet.sh
```diff
@@ -11,6 +11,21 @@ source "$SCRIPTS_DIR/lib/docker.sh"
...
+# we'd make 1 request per file (15+ *sonnet files), now we clone at most once per run.
+JSONNET_LIBS_REPO="$HOME/.outreach/.cache/jsonnet-libs"
+
+if [[ -d "$JSONNET_LIBS_REPO" ]]; then
+  pushd "$JSONNET_LIBS_REPO" || fatal "Could not find jsonnet-libs cache dir"
+  git pull
+  popd || fatal "Could not change directory out of jsonnet-libs cache dir"
+else
+  mkdir -p "$(dirname "$JSONNET_LIBS_REPO)"
+  git clone --quiet --single-branch git@github.com:getoutreach/jsonnet-libs "$JSONNET_LIBS_CACHE" >/dev/null
```
malept, line 26:
```suggestion
  git clone --quiet --single-branch git@github.com:getoutreach/jsonnet-libs "$JSONNET_LIBS_REPO" >/dev/null
```

shell/build-jsonnet.sh
```diff
@@ -24,8 +39,8 @@ email="${DEV_EMAIL:-$(git config user.email)}"
 appImageRegistry="${DEVENV_DEPLOY_IMAGE_REGISTRY:-"$(get_docker_pull_registry)"}"
 
 kubecfg \
+  --jpath "$JSONNET_LIBS_CACHE" \
```
malept, line 42:
```suggestion
  --jpath "$JSONNET_LIBS_REPO" \
```

malept (commented):
Sorry, I didn't have a fully-formed suggestion the last time as I was dealing with this along with the other unrelated issue.

shell/build-jsonnet.sh
malept: @AndrewWinterman Yeah, just don't collide with anything else and you can use `~/.outreach/.cache`.

malept (approved):

jaredallard:
You might also want to look at https://github.com/jsonnet-bundler/jsonnet-bundler longer-term, that eventually going to be a route looked at for this

PR #910, opened 2025-05-22
fix(mise): attempt multiple times to install mise itself

malept:

Sometimes downloading mise or its install script fails, we need to account for that.

## Jira ID

[DT-4766]

[DT-4766]: https://outreach-io.atlassian.net/browse/DT-4766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shell/lib/mise.sh
+++ b/shell/lib/mise.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# mise related functions. Assumes logging.sh is sourced.
+# mise related functions. Assumes logging.sh and shell.sh are sourced.
 
 # Installs `mise` if it isn't already found in PATH.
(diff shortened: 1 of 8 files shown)


shell/lib/mise.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# mise related functions. Assumes logging.sh is sourced.
+# mise related functions. Assumes logging.sh and shell.sh are sourced.
```
malept, line 3: `shell.sh` is now required for `retry`.

ekalinichev (approved):

PR #929, opened 2025-06-25
fix: use realpath instead of python

evadnoob:
use realpath, always available, instead of python

realpath in python
```
time python -c "import os,sys; print(os.path.realpath(sys.argv[1]))" service.yaml
/Users/davidboon/Projects/outreach/devbase/evadnoob_fix_get-absolute-path-remove-python-dependency/service.yaml
python -c "import os,sys; print(os.path.realpath(sys.argv[1]))" service.yaml  0.01s user 0.02s system 7% cpu 0.436 total
```

realpath from coreutils
```
time realpath service.yaml
/Users/davidboon/Projects/outreach/devbase/evadnoob_fix_get-absolute-path-remove-python-dependency/service.yaml
realpath service.yaml  0.00s user 0.00s system 15% cpu 0.013 total
```

the exec to python to noticibly slower than using realpath from
coreutils

### DIFF
--- a/scripts/devbase.sh
+++ b/scripts/devbase.sh
@@ -10,8 +10,7 @@ gojqVersion="v0.12.17"
 
 # get_absolute_path returns the absolute path of a file
 get_absolute_path() {
-  python="$(command -v python3 || command -v python)"
-  "$python" -c "import os,sys; print(os.path.realpath(sys.argv[1]))" "$1"
+  realpath "$1" 
 }
 
 # gojq returns the path to a JIT-downloaded gojq binary.


malept (changes requested):
For context:

Per [Stack Exchange](https://apple.stackexchange.com/questions/450035/is-the-unix-realpath-command-distributed-with-macos-ventura), `realpath` wasn't distributed with macOS until version 13, which probably wasn't supported by `devbase` when this implementation was written.

At a high level, this seems fine. However, a few things need to be resolved before this can be merged.

* Linting failed in CI.
* There's no reason for `get_absolute_path` to exist anymore, the one callsite can just call `realpath` directly.
* This only changes `scripts/devbase.sh`, but this file is managed by Stencil (this module, in fact). So the changes need to also happen in `templates/scripts/devbase.sh.tpl` as well.

PR #953, opened 2025-08-15
fix(bootstrap): select modules from stencil.lock correctly

malept:

Does what it says on the tin.

Also, add tests for some of the newer functions in `shell/lib/bootstrap.sh` to prevent this issue from happening for those.

### DIFF
--- a/shell/lib/bootstrap.sh
+++ b/shell/lib/bootstrap.sh
@@ -190,7 +190,7 @@ stencil_arg() {
 # from stencil.lock.
 stencil_module_version() {
   local module_name="$1"
-  "$YQ" --raw-output ".modules | select(.name == \"$module_name\").version" "$(get_stencil_lock)"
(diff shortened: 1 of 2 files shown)


shell/lib/bootstrap.sh
```diff
@@ -190,7 +190,7 @@ stencil_arg() {
 # from stencil.lock.
 stencil_module_version() {
   local module_name="$1"
-  "$YQ" --raw-output ".modules | select(.name == \"$module_name\").version" "$(get_stencil_lock)"
+  "$YQ" --raw-output ".modules[] | select(.name == \"$module_name\").version" "$(get_stencil_lock)"
```
ekalinichev, line 193: So did this just never work before?

ekalinichev (approved):

shell/lib/bootstrap.sh
malept: Correct. Fortunately it currently is only triggered by the CircleCI orb sync script, which has limited use at the moment.

PR #961, opened 2025-08-22
feat: ability of skipping e2e pipeline

honzap:

Got inspired [here](https://github.com/getoutreach/devbase/blob/main/orbs/shared/jobs/test.yaml#L43), I'd propose ability to skip e2e pipeline if someone wants.

When there are no tests in the project, the env is spin up and the machine time is then wasted (which can be >1 minute).

## Jira ID

[XX-XX]


malept:
There is already a mechanism elsewhere to do this.

PR #984, opened 2025-09-24
fix: buf version to 1.57.2

AndrewWinterman:

I'd like to test out buf in a new service (tombstone), but the existing version of buf is so old it doesn't support the version2 configuration files.

This PR upgrades the buf version.

I am still testing this out, but I think it should be safe since buf still supports v1 yaml schemas.

## Jira ID

[XX-XX]

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -12,7 +12,7 @@ gh: 2.62.0
 gojq: v0.12.17
 
 # protobuf formatters/plugins/tools/etc
-buf: 1.8.0 # formatter
+buf: 1.57.2 # formatter
 protoc-gen-validate: 0.6.7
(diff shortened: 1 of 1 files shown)


malept (changes requested):
I already discussed this with Andrew, but this needs an ecosystem regression test before we can approve.

versions.yaml
```diff
@@ -12,7 +12,7 @@ gh: 2.62.0
 gojq: v0.12.17
 
 # protobuf formatters/plugins/tools/etc
-buf: 1.8.0 # formatter
```
malept: Deleted the comment because it will be more than a formatter starting with #1009

malept (approved):

PR #1124, opened 2026-04-09
feat(ci): add separate E2E cache

malept:

PR #1117 added `~/.local/share/mise` to the shared daily cache, bloating it for every job (~500MB-1GB extra) even though only E2E jobs (machine executor) need cached mise tool installs. Non-E2E jobs run in Docker where most common tools are pre-installed.

This PR splits CircleCI caching into two parallel tracks so each executor type restores only what it can actually use:

| Cache       | Executor             | Key                                                   | Paths                                                       |
|-------------|----------------------|-------------------------------------------------------|-------------------------------------------------------------|
| Daily       | `testbed-docker-aws` | `v1-daily-cache-{{ arch }}-{{ epoch }}`               | `~/.cache`, `~/.asdf`, `~/.outreach/.cache`                 |
| E2E         | `testbed-machine`    | `v1-e2e-daily-cache-{{ arch }}-{{ epoch }}`           | `~/.local/share/mise`, `~/.cache/go-build`, `~/go/pkg/mod`  |
| Daily toggle| `testbed-docker-aws` | `v1-toggle-daily-cache-{{ arch }}-{{ .Revision }}`    | `~/cache-toggle`                                            |
| E2E toggle  | `testbed-machine`    | `v1-toggle-e2e-daily-cache-{{ arch }}-{{ .Revision }}`| `~/cache-toggle-e2e`                                        |

CircleCI scopes cache keys by executor architecture, so a cache saved from a Docker job is unreachable to a machine job. The new `save_e2e_cache` job runs on `testbed-machine` so the E2E save and restore happen under the same arch.

The `install_e2e_tools` parameter on `setup_environment` (added by #1135) now also controls E2E cache restoration: when `true`, it restores `v1-e2e-daily-cache` in addition to selecting the slim E2E mise toolset. The E2E job itself sets `restore_cache: false` because the daily cache is unreachable from machine jobs anyway.

The cache-key prefix reverts from `v2-` (introduced by #1117 to invalidate caches that lacked `~/.local/share/mise`) back to `v1-`, since `~/.local/share/mise` is no longer in the daily cache. Existing `v1-` daily caches on main are layout-compatible with this PR's daily cache.

## Jira ID

[DT-5282]

## Notes for your reviewers

- The new `save_e2e_cache` job and the E2E cache itself are transitional. They will be retired once self-hosted runners land under [DT-4971].
- Out of scope (deferred to follow-up PRs): checksum-based cache keys, full `asdf` removal.
- Repos using the `stencil-golang` Stencil module already provide a `Makefile` `build` target via `devbase`, which the new `save_e2e_cache` job uses to populate the E2E cache.

[DT-4971]: https://outreach-io.atlassian.net/browse/DT-4971
[DT-5282]: https://outreach-io.atlassian.net/browse/DT-5282


marnagy (approved):
Since we will be reverting the daily cache key, is it possible that there is OLD cache in CI that might get used between us releasing `rc` version of this repo and some people running their pipelines?
I'm not yet fully sure how the cache in CI works.

palootcenas-outreach:
I tested this on an internal repo ([detailed results](https://github.com/getoutreach/gearbox/pull/1810#issuecomment-4214117526)) and ran into a few problems.

**All jobs run `mise install`, not just e2e**

Docker images don't ship repo-specific tool versions. Every job runs `mise install` (and `asdf install` where applicable) during `setup_environment`. Without cached `~/.local/share/mise` and `~/.asdf`, Docker jobs reinstall everything from scratch: 204s setup vs 8.7s with #1117. Saving 73s on a smaller cache restore doesn't offset a 195s setup regression.

**The e2e cache is unreachable by e2e jobs**

`save_cache` runs on Docker where `{{ arch }}` evaluates to `arch1-linux-amd64-6_85`. E2e runs on a machine executor where it's `arch1-linux-amd64-6_106`. These never match, so e2e jobs can't find the cache at all. The split doesn't help them either.

**`~/.asdf` has been in the daily cache since it was introduced**

This PR removes it for all non-e2e jobs. That's a regression from even the pre-#1117 state.

**Numbers**

Total stored: 3.3 GiB (1.3 + 2.0) vs 3.2 GiB with #1117. Docker jobs go from 195s to 608s. E2e jobs get nothing due to the arch mismatch.

I think #1117's single cache approach is simpler and works. If cache size near the 3 GiB limit becomes a concern, keying on `checksum("mise.toml")-checksum("mise.lock")` instead of `{{ epoch }}` would deduplicate entries without removing content that all jobs depend on.

malept:
> Since we will be reverting the daily cache key, is it possible that there is OLD cache in CI that might get used between us releasing `rc` version of this repo and some people running their pipelines?

Those caches will age out via CircleCI's normal eviction, no manual cleanup is needed.

malept:
> **All jobs run `mise install`, not just e2e**

* Docker jobs use image-based caching now (https://github.com/getoutreach/devbase/pull/1132)
* The number of repos which use custom tools defined in `mise` is limited. Even when you take into consideration the custom tools defined in `.tool-versions` when we migrate `asdf` to `mise` (finally), it won't affect the GitHub API calls as much as all of the default tools from `devbase`.

> **The e2e cache is unreachable by e2e jobs**

A separate e2e cache was created in this PR.
 
> **`~/.asdf` has been in the daily cache since it was introduced**

Re-added for the time being. It's still going away in a follow-up PR.

PR #1147, opened 2026-05-18
fix: use first tool version from .tool-versions when multiple are declared

marnagy:

What is says on the tin.

## Jira ID

[DT-5340](https://outreach-io.atlassian.net/browse/DT-5340)

## Notes for your reviewers

[DT-5340]: https://outreach-io.atlassian.net/browse/DT-5340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
Rovo Dev code review: <strong>Rovo Dev has reviewed this pull request</strong>
Any suggestions or improvements have been posted as pull request comments.


malept (changes requested):
~~The design of this change is approaching the problem from the wrong angle. Try this instead:~~

[REMOVED, see [later comment](https://github.com/getoutreach/devbase/pull/1147#pullrequestreview-4311879376)]

~~Also, this fix is missing tests (originally for `version_all_from_toolversions` but for the suggestion, `versions_from_toolversions`).~~

malept (changes requested):
That being said, I think the simpler solution is to just install the first Node.js version and ignore the rest. Still needs regression tests though.

malept (changes requested):
Remove version_all_from_toolversions, it's unused

malept (changes requested):
1. Revert changes in `shell/ci/env/mise.sh` and `shell/circleci/machine.sh`, it's not worth changing the order.
2. Fix function comment for `version_from_toolversions` - it's referencing a now-deleted function and the actual change can be better integrated into the doc comment, rather than appending a note at the end.
3. Add regression tests for `version_from_toolversions`.

malept:
Also probably needs a better PR title

shell/lib/mise_test.bats
```diff
@@ -0,0 +1,66 @@
...
+
+setup() {
+  REPO_DIR="$(mktempdir devbase-mise-test-XXXXXX)"
+}
+
+teardown() {
+  rm -rf "$REPO_DIR"
+}
+
+# Helper: write a .tool-versions file in $REPO_DIR
```
malept: suggestion: I generally don't expect folks to know the `printf` args, so the docs should be clear about what's happening.

```suggestion
# write_tool_versions [entry...]
#
# Write a .tool-versions file in $REPO_DIR with one or more entries.
```

shell/lib/mise_test.bats
```diff
@@ -0,0 +1,66 @@
...
+write_tool_versions() {
+  printf '%s\n' "$@" >"$REPO_DIR/.tool-versions"
+}
+
+# --- version_from_toolversions ---
+
```
malept: suggestion: unnecessary comment, as all of your tests are prefixed
```suggestion
```

shell/lib/mise_test.bats
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bats
+
+bats_load_library "bats-support/load.bash"
+bats_load_library "bats-assert/load.bash"
+
+load logging.sh
+load shell.sh
```
malept: question: are these needed for the tests to run?

malept (changes requested):

shell/lib/mise_test.bats
marnagy: Removed in later commit

malept (approved):

PR #1161, opened 2026-07-10
fix(kubeconform): validate manifests from local schema caches to avoid GitHub rate limiting

malept:

The `kubecfg` linter runs `kubeconform`, which downloaded Kubernetes and CRD JSON schemas from `raw.githubusercontent.com` at validation time. GitHub has been aggressively rate-limiting that host, so CI linting and release jobs intermittently failed with HTTP 429. Those URLs are unauthenticated and cannot be authenticated, so the fix is to stop fetching schemas over the network just-in-time.

`kubeconform` now validates against local git clones of the schema repos, cached under `~/.outreach/.cache` (persisted by the existing CircleCI cache, so no CI image changes are needed).

## Jira ID

[DT-5424]

## Notes for your reviewers

- The offline guarantee comes from the wrapper emitting only local filesystem schema locations: never `-schema-location default` or an `http(s)` URL. Because a single HTTP 429 from any network location aborts `kubeconform`'s entire run, reintroducing one would defeat the fix. This was confirmed with `strace` (zero network syscalls once the cache exists).
- `kubernetes-json-schema` is multiple GB when checked out in full, so it is cloned shallow, blobless, and sparse to only the directories for the pinned Kubernetes version.
- I took this opportunity to upgrade the used `kubeconform` from `0.6.4` to `0.8.0` to validate that the bug wasn't just in the old version we were using.
- `test-macos` failure is addressed in a different PR (https://github.com/getoutreach/devbase/pull/1160)

### TODO

- [x] Merge #1160 
- [x] Merge main into this PR
- [x] Wait for CI to pass

[DT-5424]: https://outreach-io.atlassian.net/browse/DT-5424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shell/lib/git_cache.sh
```diff
@@ -0,0 +1,72 @@
...
+    cacheDir="$DEVBASE_CACHE_DIR/$cacheBasename"
+  fi
+
+  if [[ -d $cacheDir ]] && git -C "$cacheDir" rev-parse --git-dir >/dev/null 2>&1; then
+    info_sub "Updating local cache" >&2
+    # A transient fetch failure should not be fatal: a usable checkout
+    # already exists, so warn and fall back to it rather than aborting.
+    if ! { git -C "$cacheDir" fetch --depth 1 &&
+      git -C "$cacheDir" reset --hard -q origin/HEAD; }; then
+      warn "Could not refresh cache at $cacheDir; using the existing checkout" >&2
```
iamrz1, line 56: **Non-blocking suggestion.** Tolerating a warm-fetch failure is reasonable, but the function returns `$cacheDir` unconditionally afterward without checking that the requested sparse/version directories are actually present. Because the caller (`kubeconform.sh`) runs kubeconform with `-ignore-missing-schemas`, a returned-but-incomplete cache can produce a "silent green": validation is skipped and the run still exits 0.

Concrete scenario where this bites:
1. `$cacheDir` exists and passes the `git rev-parse` check on line 42, so we take this warm-cache branch.
2. `git fetch` / `git reset --hard origin/HEAD` fails or leaves the tree incomplete (e.g. corrupted ref/tree, interrupted prior reset, or the requested `v${k8sVersion}-standalone` dir simply isn't materialized).
3. We `warn` and fall through; `echo "$cacheDir"` returns the dir as if healthy.
4. kubeconform finds no schema files under that path and, thanks to `-ignore-missing-schemas`, skips those resources instead of failing — reporting success while validating nothing.

Note the failure is narrow: an intact prior checkout after a plain network blip is fine (schemas are still there from last time). The risky case is specifically a fetch/reset failure that leaves the expected version directories absent.

Suggested hardening (optional): after a tolerated refresh failure, verify the requested `sparsePaths` still exist on disk; if any are missing, treat it as a cache miss (`rm -rf` + re-clone) rather than returning a dir that yields a false pass.

iamrz1 (approved):

shell/lib/git_cache.sh
malept: Fixed in d34cef1. After a tolerated refresh failure, the warm path verifies every requested sparse path is present and non-empty and calls fatal otherwise.

Not implemented as rm -rf + re-clone: the check only fires when the refresh already failed (remote unreachable), so a re-clone fails too and would return an empty dir with exit 0 — the same silent-green. When the remote is reachable, git reset --hard / sparse-checkout set already re-materialize the paths and the check passes. Verified against the real yannh/kubernetes-json-schema remote: with the promisor unreachable, git sparse-checkout set exits 0 while leaving files unmaterialized. Regression test added in git_cache_test.bats.

----
:robot: _Comment generated with AI_

PR #1167, opened 2026-07-16
fix(e2e): bump devenv to 1.65.0

iamrz1:

Bumps the e2e-pinned `devenv` version from **1.64.7 → 1.65.0** (in `mise.e2e.toml`, with `mise.e2e.lock` updated to match).

The shared orb's `e2e` job installs `devenv` from the version pinned in `mise.e2e.toml` (via mise; `devenv_pre_release` defaults to `false`). That pin was still `1.64.7`, which predates the Velero **node-agent memory fix** released in devenv `v1.65.0`. As a result, every consumer's e2e job (e.g. `searchproxy`) was still running `1.64.7` and intermittently failing while waiting for restic (fs-backup) restores to complete:

```
failed to wait for restic restores to complete: timed out waiting for restic restores to complete:
[admingw--bento1a/redis-0 kafka/strimzi-topic-operator-... orgservice--bento1a/redis-0 searchstore--bento1a/redis-0]
```

Bumping the pin lets the released fix actually reach e2e once a new orb is cut and adopted.

## Jira ID

[XX-XX]

## Notes for your reviewers

- **Lockfile verified:** all 6 `devenv` platform entries (linux-arm64 (+musl), linux-x64 (+musl), macos-arm64, macos-x64) were cross-checked against the live `v1.65.0` GitHub release — URLs, sha256 checksums, and numeric asset IDs all match the release `checksums.txt`/asset digests. The `-musl` variants intentionally reuse the non-musl tarball, matching the prior `1.64.7` convention.
- **Follow-up required for the fix to land:** this bump alone is not sufficient — a new `shared` orb release must be cut and consumers (searchproxy, etc.) must adopt it.
- **Possible cache staleness:** the E2E daily cache (`~/.local/share/mise`) is restored via a time-based prefix key rather than a lockfile hash, so the first runs after merge may still restore a cache containing the old `1.64.7` binary until the daily cache rotates. mise should re-resolve to `1.65.0` from the updated lock, but flagging in case a forced cache-bust is desired.


malept (approved):
The fix is fine, the PR description is something like a dozen lines too long.

PR #1177, opened 2026-07-29
fix: add stencil upgrade to stencil:upgrade

marnagy:

When restenciling, I noticed that the binary installed `stencil` and mise-managed `stencil` have different versions due to `stencil:upgrade` does not include upgrade of `stencil`. This PR should fix it.

## Jira ID

[XX-XX]

### DIFF
--- a/.mise/tasks/stencil/upgrade
+++ b/.mise/tasks/stencil/upgrade
@@ -4,4 +4,5 @@
 #MISE depends=["devbase:sync"]
 #MISE depends_post=["stencil:post"]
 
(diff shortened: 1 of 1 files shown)


iamrz1 (approved):

malept:
We shouldn't be doing this automatically, this should be part of the developer laptop setup.

marnagy:
@malept 
Before running `stencil` through `mise`, automatic self-update was part of the stencil. I assumed we want to preserve this. Why the change?

PR #1182, opened 2026-08-11
test(github): stop the pre-release install test failing after a stable release

malept:

`latest_github_release_version` only drops `--exclude-pre-releases` when pre-releases are requested, so it returns the newest release of any kind. Between a stable stencil release and the next commit that re-creates that repo's `unstable` release, the newest release is a stable one, and the test's assertion that the installed version matches `(rc|unstable)` failed.

Cover the flags the function builds with a stubbed `gh`, so that coverage no longer depends on another repo's release state or on the GitHub API rate limit. Skip the live install test while the newest release is stable. Correct the doc comments, which described the stricter behavior the test assumed.

## Notes for your reviewers

The first commit is almost all relocation. `stub_command` and `assert_stub_not_called` move from `metrics_test.bats` to `test_helper.sh` byte-identical, and only `stub_argv` is new.

The skip guard spends one extra `gh release list` call. Reading the tag from the install log would avoid it, but `(rc|unstable)` also matches CI paths such as `/home/circleci/...`.

---
:robot: _Generated by [Claude Code](https://claude.ai/code/session_01NWz3v5hdZRcyB6RLPzDgGZ)_

### DIFF
--- a/shell/lib/github_test.bats
+++ b/shell/lib/github_test.bats
@@ -23,6 +23,8 @@ setup() {
     echo 'github-cli = "latest"'
     echo 'wait-for-gh-rate-limit = "1.1.1"'
   } >>"$MISE_GLOBAL_CONFIG_FILE"
+
+  setup_command_stubs
 }
 
 teardown() {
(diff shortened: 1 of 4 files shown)


shell/lib/github_test.bats
```diff
@@ -35,13 +37,48 @@ teardown() {
...
+  stub_command gh "v1.45.0"
+
+  run latest_github_release_version getoutreach/stencil true
+  assert_success
+  assert_output "v1.45.0"
+
+  run stub_argv gh
+  assert_line "--exclude-drafts"
+  refute_line "--exclude-pre-releases"
+}
```
marnagy, line 72: Nitpick: This test only asserts `--exclude-drafts`/`--exclude-pre-releases`, not the repo slug like the test above it. Not a real gap — just flagging for symmetry, feel free to ignore.

- Generated using AI

marnagy (approved):
LGTM — verified build/lint/tests pass locally (13/13 bats, incl. live GitHub tests). Fix correctly targets the newest-release-vs-pre-release selection bug; new stub-based tests decouple coverage from live GitHub state.

- Generated using AI

shell/lib/github_test.bats
malept: The split is deliberate. The test above pins the full argument shape, including the slug, and this one pins only what differs between the two modes. Repeating the slug assertion here would not catch anything the other test misses.

---
_Generated by [Claude Code](https://claude.ai/code)_

